PR #8, opened 2020-09-23
New function to look up timezones in database

fiffty-50:


fiffty-50:
New function to look up timezones in database

PR #12, opened 2020-11-29
Entry now inherits from QObject

fiffty-50:
The Entry class now inherits from QObject. This has 2 key advantages:
1) Qt can manage our memory if we declare a parent in the constructor of the Entry Object
2) Entry objects can emit signals (e.g. commitSuccessful or sqlError(QString&)

There are also new constructors for an Entry object, which I think will make using the class more user-friendly. Instead of using QString tableName, we can now use an enum Db::table

Example for creating an entry object for a flight at row 34 in the database:

Entry sampleFlight = Entry(Db::flights, 34, this);

I believe with these changes, using subclasses for Pilots, Aircraft and Flights is unneeded bloat and we could remove those classes and use an Entry(Db::flights, int rowID) constructor instead of Flight(rowID). The subclasses as of now don't really do anything else rather than providing a named constructor anyway.

### DIFF
--- a/mainwindow.cpp
+++ b/mainwindow.cpp
@@ -125,7 +125,7 @@ void MainWindow::on_actionAircraft_triggered()
 
 void MainWindow::on_actionNewFlight_triggered()
 {
-    NewFlightDialog nf = NewFlightDialog(this, Db::createNew);
+    NewFlightDialog nf = NewFlightDialog(Db::createNew, this);
     nf.exec();
 
 }
--- a/src/classes/aircraft.cpp
+++ b/src/classes/aircraft.cpp
@@ -19,13 +19,14 @@
 #include "debug.h"
 
 
-Aircraft::Aircraft()
+Aircraft::Aircraft(QObject *parent)
 {
-
+    this->setParent(parent);
 }
 
-Aircraft::Aircraft(int tail_id)
+Aircraft::Aircraft(int tail_id, QObject *parent)
 {
+    this->setParent(parent);
     //retreive database layout
     const auto dbContent = DbInfo();
     auto table = QLatin1String("tails");
@@ -61,13 +62,14 @@ Aircraft::Aircraft(int tail_id)
     }
 }
 
-Aircraft::Aircraft(QMap<QString, QString> newData)
+Aircraft::Aircraft(QMap<QString, QString> newData, QObject *parent)
 {
+    this->setParent(parent);
     QString table = "tails";
 
     //retreive database layout
     const auto dbContent = DbInfo();
-    columns = dbContent.format.value(table);
+    auto columns = dbContent.format.value(table);
     //Check validity of newData
     QVector<QString> badkeys;
     QMap<QString, QString>::iterator i;
--- a/src/classes/aircraft.h
+++ b/src/classes/aircraft.h
@@ -26,11 +26,11 @@
  */
 class Aircraft : public Entry
 {
-//    using Entry::Entry;
+Q_OBJECT
 public:
-    Aircraft();
-    Aircraft(int tail_id);
-    Aircraft(QMap<QString, QString> newData);
+    Aircraft(QObject* parent = nullptr);
+    Aircraft(int tail_id, QObject* parent = nullptr);
+    Aircraft(QMap<QString, QString> newData, QObject* parent = nullptr);
 };
 
 #endif // AIRCRAFT_H
--- a/src/classes/flight.cpp
+++ b/src/classes/flight.cpp
@@ -18,13 +18,14 @@
 #include "flight.h"
 #include "debug.h"
 
-Flight::Flight()
+Flight::Flight(QObject *parent)
 {
-
+    this->setParent(parent);
 }
 
-Flight::Flight(int flight_id)
+Flight::Flight(int flight_id, QObject *parent)
 {
+    this->setParent(parent);
     //retreive database layout
     const auto dbContent = DbInfo();
 
@@ -62,13 +63,14 @@ Flight::Flight(int flight_id)
     }
 }
 
-Flight::Flight(QMap<QString, QString> newData)
+Flight::Flight(QMap<QString, QString> newData, QObject *parent)
 {
+    this->setParent(parent);
     QString table = "flights";
 
     //retreive database layout
     const auto dbContent = DbInfo();
-    columns = dbContent.format.value(table);
+    auto columns = dbContent.format.value(table);
 
     //Check validity of newData
     QVector<QString> badkeys;
--- a/src/classes/flight.h
+++ b/src/classes/flight.h
@@ -26,11 +26,11 @@
 
 class Flight : public Entry
 {
-//    using Entry::Entry;
+Q_OBJECT
 public:
-    Flight();
-    Flight(int flight_id);
-    Flight(QMap<QString, QString> newData);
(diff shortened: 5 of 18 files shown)


KotzaBoss:
### TL;DR
Agree on QObject, Disagree on removing subclasses: Should create a better defined hierarchy of abstract base and concrete derived classes, Decouple database operations from Entry and exploit polymorphism.

---

I see the potential of "linking" Entry with QObject it would indeed add some functionality and memory management. While further testing is needed i disagree with the last part of the request. The whole idea of having separete Flight, Pilot etc classes is exactly the reason we should have a robust base Entry in the first place. Consider the following trivialish example

```c++
class Entry { ... };  // Can i assume that all entries do the same exact database validation and operations?

void Db::add_pilot(Entry e);  // perhaps a specific method of Entry is called depending on the function?
void Db::add_flight(Entry e, Operation op); // Or perhaps hint to it with a separete type??? 
void Db::find_acft(Entry e);

// However subclassing the Entry class can actually lead to an interesting result
class Entry {
  bool do() = 0;  // Job of the subclasses to determine what must be done with their encapsulated data 
};

class Pilot : Entry {
  Data CUSTOM_DATA
  bool do(){
    // do stuff with CUSTOM DATA
  }
};

// if we call a more generic add_entry function with a Pilot,  Flight, ... class,
// through polymorphism we give the responsibility
// to the subclasses on what do() does,
// delegating the responsibility to the subclasses and not a monolithic mega class
void Db::add_entry(Entry* e) {
    ...
    e->do();
   ...
}
```
IMO good code isnt only about blindly reducing LOC. Sometimes a little more lines
of specific well encapsulated data and operations is so much more "cost-efficient"
in the long term. Bloat can come in different ways. Some time literally in LOC
some times in stuffing a class with too much responsibily.

To finish off with your own example. 
> I believe with these changes, using subclasses for Pilots, Aircraft and Flights is unneeded bloat and we could remove those
> classes and use an Entry(Db::flights, int rowID) constructor instead of Flight(rowID).

I would prefer to write `Flight(rowID)` over `Entry(Db::flights, int rowID)` any day. It is more explicit while being less to write.

### Proposal
Due to the work that you did in the entry class, moving the responsibility to the subclasses would require simple finding a clean
way to abstract the possible operations a `Entry` object would demand form the database.
Instead of relying on the `Entry` object it self to do operations in the db **from itself**, we can use polymorphism in the `Db` api to use the encapsulated data of the entries to do the operations, **at the database level**. The `Entry` objects (and of course the subclasses) would only act as packets of information. This is a rough idea of how the database-entry relation could go. If it sounds good i can shift my attention to that instead of the NewPilot functionality.
I believe i can be of more use there.

### PS
Off course the database api could just have an overload for Flight, Acft, Pilot ... but i believe that polymorphism is 
a more clean way to do it IMO. Of course this is debateable and not the really the point of this merge request.

fiffty-50:
Entry::commit is 'smart' enough to do a number of things, for which we would have to write a lot of database functions. For example the hypothetical functions

```cpp
Db::addPilot();
Db::editPilot();
Db::addAircraft();
Db::editTail();
Db::addTail();
Db::addFlight();
Db::editFlight();
```


can all be replaced with Entry::commit, since it checks whether an entry exists and calls the internal Entry::commit and Entry::update functions accordingly. Also, thanks to its 'position' member variable it knows its place in the database (Table and row).

We can keep the subclasses around, the reason I was happy to remove them was because they don't do anything specific really, except for being easier constructors for the entry class. On a database level, it is irrelevant whether we are adding a flight, a pilot or a tail, the code will be the same. All that changes is the names of the columns and the name of the table the data is inserted to. I couldn't think of any tasks specific to a tail entry that is different from a pilot entry for example, but if you have some ideas in that direction, let's explore it!

Moving the database functionality out of the Entry class would be as trivial as transferring these member functions:
```cpp
Entry::commit();
Entry::insert();
Entry::udpate();
Entry::remove();
Entry::exists();
```

As for an entry being a QObject, even if it doesn't emit signals I would still go for it considering we can assign parents and let qt manage the memory this way.

KotzaBoss:
A great code goal is to have self documenting code. Its ok to have small seemingly simple classes that only helps with the naming. its better to see
```c++
QuantumResult universe_breaking_function(PlaceInUniverse piu);
```
than
```c++
string universe_breaking_function(int i);
```
Even if `QuantumResult` is just a simple function that inherits from `string` and just adds a small parsing function or
`PlaceInUniverse` is literally an int. It is documenting itself.
What i propose does more than that of course but the point stands. Explicit is better than implicit after all
python example:
```python
# perfectly good python code to create a simple custom exception
# doesnt really do anything is but document itself the functionality is inherited (its ok!)
class UnrecoverableState(Exception):
  pass
```

My approach is about the fundamental flow and coupling of the code. My idea is that an entry doesnt know anything besides the imformation it holds it doesnt care about the database it will end up in.
It is the database's job to commit, find, change ... the entries passed.
Therefore from what you wrote the change is from
```c++
Entry::commit();
Entry::insert();
Entry::udpate();
Entry::remove();
Entry::exists();
```
to
```c++
Db::commit(Entry* e);
Db::insert(Entry* e);
Db::udpate(Entry* e);
Db::remove(Entry* e);
Db::exists(Entry* e);
```
Even if we dont take advantage of polymophism, overloading db functions is still preferable to embedding database functionality in entries.
Now we have entries that hold relevant data and a database api that knows what todo with the data given.
> On a database level, it is irrelevant whether we are adding a flight, a pilot or a tail, the code will be the same

That simplifies the db api even further. 
> I couldn't think of any tasks specific to a tail entry that is different from a pilot entry for example

Even if we do, extending the operations will be much easier since we would update our db api to take into account the intricacies of the new Entry subclass. While still maintaining in-dependancy between entries and the database.
Let me reiterate that my idea is about the overall architecture. Entries are data, Database operates **on** data.
I can work in it on a branch. Even if we reject it the code can speak for itself.

To quote Bjarne Stroustrup when he was asked what is something you are actively doing these days on your code.
> [...] im removing the clever bits [...]

Classes should be smart enough not smarter or dumber. :D 

For the QObject. I dont imidiatelly see anything wrong besides bloating the classes with functions they propably wont ever use. If we design the classes well they will clean their own mess well enough. Refactoring that is at this point more boring than anything i think.

fiffty-50:
Okay I see your point. Let's keep the entry subclasses around then even if they're without much functionality for now. 

One thing I'd like to address before you start working on it is a naming ambiguity I would like to change. It has to do with the distinction between an aircraft and a tail. If you look at for example the newtail dialog, you'll see that the Aircraft Class is used to interact with the tails database whereas the aircraft database is used with a generic Entry object. This can be confusing and I think we should make it clearer which is which. Since the entry class can "clean" itself from columns that don't concern tables they are referencing it is possible to use it the way it is, but I think we can do better, here is my suggestion:

For each table in the database, there should be an entry-subclass, i.e.
Pilots, Tails, Flights, Airports and Aircraft.

When referencing an **individual aircraft** (D-ABCD), int the **tails** table the **tail** class should be used, when referencing a **Type** (Boeing 737-800), the **aircraft** class, referencing the aircraft table should be used.


As for the QObject, with Entry being a data container type there is no need for it to inherit from QObject, but in that case the database should be, I have a couple of ideas where it is very useful to have a Signal emitted upon let's say commit successful instead of just a 'True' return code.

PR #13, opened 2020-12-06
implementation of init

fiffty-50:
init creates a connection and fills in the member variables of the database class.

### DIFF
--- a/src/experimental/Db.cpp
+++ b/src/experimental/Db.cpp
@@ -1,0 +1,42 @@
+#include "Db.h"
+
+using namespace experimental;
+
+bool DB::init()
+{
+    const QString driver("QSQLITE");
+
+    if (QSqlDatabase::isDriverAvailable(driver)) {
+
+        QDir directory("data");
+        QString databaseLocation = directory.filePath("logbook.db");
+        QSqlDatabase db = QSqlDatabase::addDatabase(driver);
+        db.setDatabaseName(databaseLocation);
+
+        if (!db.open()) {
+            DEB("DatabaseConnect - ERROR: " << db.lastError().text());
+            return false;
+        } else {
+            DEB("Database connection established.");
+            // Enable foreign key restrictions
+            QSqlQuery query("PRAGMA foreign_keys = ON;");
+            // Retreive database layout and store in member variables
+            tables = db.tables();
+
+            QStringList columnNames;
+            for (const auto &table : tables) {
+                columnNames.clear();
+                QSqlRecord fields = db.record(table);
+                for (int i = 0; i < fields.count(); i++) {
+                    columnNames.append(fields.field(i).name());
+                    database_layout.insert(table, columnNames);
+                    ///[F] database_layout.value("pilots") gives you the columns for the pilots table
+                }
+            }
+            return true;
+        }
+    } else {
+        DEB("DatabaseConnect - ERROR: no driver " << driver << " available");
+        return false;
+    }
+}
--- a/src/experimental/Db.cpp
+++ b/src/experimental/Db.cpp
@@ -6,37 +6,29 @@ bool DB::init()
 {
     const QString driver("QSQLITE");
 
-    if (QSqlDatabase::isDriverAvailable(driver)) {
+    if (!QSqlDatabase::isDriverAvailable(driver)) return false;
 
-        QDir directory("data");
-        QString databaseLocation = directory.filePath("logbook.db");
-        QSqlDatabase db = QSqlDatabase::addDatabase(driver);
-        db.setDatabaseName(databaseLocation);
+    QDir directory("data");
+    QString databaseLocation = directory.filePath("logbook.db");
+    QSqlDatabase db = QSqlDatabase::addDatabase(driver);
+    db.setDatabaseName(databaseLocation);
 
-        if (!db.open()) {
-            DEB("DatabaseConnect - ERROR: " << db.lastError().text());
-            return false;
-        } else {
-            DEB("Database connection established.");
-            // Enable foreign key restrictions
-            QSqlQuery query("PRAGMA foreign_keys = ON;");
-            // Retreive database layout and store in member variables
-            tables = db.tables();
+    if (!db.open()) return false;
 
-            QStringList columnNames;
-            for (const auto &table : tables) {
-                columnNames.clear();
-                QSqlRecord fields = db.record(table);
-                for (int i = 0; i < fields.count(); i++) {
-                    columnNames.append(fields.field(i).name());
-                    database_layout.insert(table, columnNames);
-                    ///[F] database_layout.value("pilots") gives you the columns for the pilots table
-                }
-            }
-            return true;
+    DEB("Database connection established.");
+    // Enable foreign key restrictions
+    QSqlQuery query("PRAGMA foreign_keys = ON;");
+    // Retreive database layout and store in member variables
+    tableNames = db.tables();
+
+    QStringList columnNames;
+    for (const auto &table : tableNames) {
+        columnNames.clear();
+        QSqlRecord fields = db.record(table);
+        for (int i = 0; i < fields.count(); i++) {
+            columnNames.append(fields.field(i).name());
+            tableColumns.insert(table, columnNames);
         }
-    } else {
-        DEB("DatabaseConnect - ERROR: no driver " << driver << " available");
-        return false;
     }
+    return true;
 }
--- a/src/experimental/Db.h
+++ b/src/experimental/Db.h
@@ -157,35 +157,6 @@ public:
             return false;
         }
     }
-
-    /*!
-     * \brief Verify entry data is sane for the database.
-     */
-    static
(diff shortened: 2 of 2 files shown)


src/experimental/Db.h
```diff
@@ -91,7 +91,8 @@ bool insertPilot(UserInput& uin)
  */
 class DB {
 private:
-    static QVector<QString> columns;
+    static QMap<QString, QStringList> database_layout; // contains the column names for all tables.
+    static QStringList tables; //contains the table names
```
KotzaBoss, line 95: Embed comment in variable name
```c++
static QStringList table_names;
```

src/experimental/Db.h
```diff
@@ -178,6 +177,7 @@ class DB {
         }
         entry.populate_data(data);
         return update(entry);
+        ///[F] maybe this should be the other way around, i.e. update calls verify instead of verify calling update?
```
KotzaBoss, line 180: Excellent suggestion! Update will be called from the gui anyway

src/experimental/Db.cpp
```diff
@@ -0,0 +1,42 @@
...
+                    ///[F] database_layout.value("pilots") gives you the columns for the pilots table
+                }
+            }
+            return true;
+        }
+    } else {
+        DEB("DatabaseConnect - ERROR: no driver " << driver << " available");
+        return false;
+    }
+}
```
KotzaBoss, line 42: Broader programming comment. There is a specific way of handling exiting state in such if/else type of logic. in order to avoid nesting if and indenting you put the failure state first as follows
```c++
if(!desirable_state) return false;
// continuing normally
if(!second_desirable_statate) return true;
// more logic
// FINALLY
return true
```
So you make your logic more linear with exit hatches as it goes further down. 
For our example it could be.
```c++
if(!availlableDriver()) return false;
auto db = QSqlDatabase(...);
if(!db.open()) return false;
setupTables(db);
return true;
```
That way the flow is up down linearly with no indentation

KotzaBoss (changes requested):
tl;dr: switch the verify-update flow, cleanup a variable name, linearise init flow (unindent the logic) and we are go for merging. 
Nice work!

fiffty-50:
changed control flow of init and renamed DB member variables for clarity.

Also added an overview of planned DB functionality for discussion. Thank you for your comments!

src/experimental/Db.h
```diff
@@ -157,18 +162,18 @@ class DB {
      * \brief Verify entry data is sane for the database.
      */
     static
-    bool verify(Entry entry)
+    Entry prepareDataForCommit(Entry entry)
+    /// [F] this function does not really verify data, it just removes bad entries,
+    /// maybe returning the 'cleaned up' entry would be better?
```
KotzaBoss, line 167: Im guessing this is an internal function to cleanup the to-be-updated entry.
However where does the "commit" come in? Is it going to just be a wrapper around insert and update?
If so what will be checked for whether Entry is to be updated or inserted?

KotzaBoss (commented):
The prepareDataForCommit bring up the issue of naming and interface.
Check comment on Db.h file.

src/experimental/Db.h
fiffty-50: That's correct. I would implement commit something like this (once prepareDataForCommit is finished):
```cpp
bool commit(Entry entry)
{
    if (exists(entry)) {
        return update(entry);
    } else {
        return insert(entry);
    }
}
```
exists would check the position member of the entry object. The default position for a new object is {"table",0}, whereas an existing entry has a rowId != 0. However, if not 0, we could (should?) also confirm the entry exists in the database. This adds some overhead but might be worth it. If we ommit this check, exists would only have to check position.second

```cpp
bool exists(Entry entry)
{
    if (entry.position.second == 0)
        return false;
    
    //Check database for row id
    QString statement = "SELECT COUNT(*) FROM " + entry.position.first +
                        " WHERE _rowid_=" + QString::number(entry.position.second);
    //this returns either 1 or 0 since row ids are unique
    QSqlQuery q(statement);
    q.next();
    int rowId = q.value(0).toInt();
    if (rowId) {
        DEB("Entry exists with row ID: " << rowId);
        return true;
    } else {
        DEB("Entry does not exist.");
        return false;
    }
}
```

KotzaBoss: Fair enough this makes sense but api wise how will it work?
For example the NewPilotDialog. what will it call?
If the user creates new entry its easy bcause we create stuff from the UserInput
but if he updates an entry? Arent we supposed to just call the update after cleaning the entry?

fiffty-50: Well, the 'clean-up' function would be called before updating or inserting, since it has to be called for both new entries as well as edited entries maybe we can call it in the constructor for entry?

The flow could look something like:

NewPilot Dialog is ready and calls
```cpp
UserInput rawData =  collectInput(QList<QLineEdit*>);

Entry toDatabase = Entry(rawData);
// constructor of entry reads metadata and calls clean up function

if (editRole == Db::editExisting)
    toDatabase.setPosition(oldPosition)
// set position if editing a new entry, otherwise default position will be 0

DB::commit(toDatabase);
// depending on position data either calls insert or update

// on success, accept dialog, otherwise prompt user.
```
What do you think?

KotzaBoss: Im not convinced this is a good course of action. It feels like db logic is spilling out again. If the user is editing an entry then simply by clicking it from the gui we have the position in the database so we can directly call `update(updated_entry)` (and being explicit in our intent). I think im missing something. 

We need to clarify this distiction of updating vs inserting not only here in db but also gui level. If by clicking the existing pilots the user is selecting Entries then the dialog should edit that `Entry` and pass it to the database via `update(Entry)`.
We could keep an `Entry` encapsulated in the dialog. If we call the overloaded constructor then we keep the entry which gets updated after the user is done editing and pass that to the `db::update(Entry)`. We can keep the edit role if you think it makes the potential outcome of the dialog explicit (which i would agree with)
```c++
class NewPilotDialog {
    ...
    Entry potential_entry;
    ...
};

void NewPilotDialog::submitForm()
{
    switch (role) {
    case Db::createNew:
        if(!DB::insert(uin)) emit QDialog::rejected();
        break;
    case Db::editExisting:
        if(!DB::update(potential_entry)) emit QDialog::rejected();
        break;
    }
}
```
It needs a bit of refactoring but i think it still keeps what we are trying to achieve with cleaning up the responsibilities

fiffty-50: In this case, what would be the distinction between UserInput and Entry? Wasn't the idea to have the entry represent a 'sane' database entry and the userInput as a 'raw' data that needs some processing before it is being fed to the database?

Keeping the old entry around in the dialog would not fix the need for cleanup either. This step is not unique to a new entry. Let me  give you an example:
Data is collected from the dialog like this:
```cpp
void NewPilotDialog::submitForm()
{
...
    auto line_edits = parent()->findChildren<QLineEdit *>();

    for (const auto &le : line_edits) {
        QString key = le->objectName();
        key.chop(8);//remove "LineEdit"
        QString value = le->text();
        if (!key.isEmpty()) {
            newData.insert(key, value);
            DEB("Data entry: " << key << value);
        }
    }
...
}
```
This gives us the following data:
```cpp
void NewPilotDialog::submitForm() 	 Creating Database Object...
void NewPilotDialog::submitForm() 	 Data entry:  "search" ""
void NewPilotDialog::submitForm() 	 Data entry:  "picfirstname" "Adam"
void NewPilotDialog::submitForm() 	 Data entry:  "piclastname" "Knox"
void NewPilotDialog::submitForm() 	 Data entry:  "company" "Lufthansa"
void NewPilotDialog::submitForm() 	 Data entry:  "alias" ""
void NewPilotDialog::submitForm() 	 Data entry:  "employeeid" ""
void NewPilotDialog::submitForm() 	 Data entry:  "phone" ""
void NewPilotDialog::submitForm() 	 Data entry:  "email" ""
````
In this example, cleanUp would remove the entry **{"search" ""}**, as it originates from searchLineEdit and is not part of the database entry.

Now, if we would use this data as is (without cleaning it up) on an 'old' entry with entry.setData(newData), the entry would be polluted and the update query would fail.

Comment (KotzaBoss) [reply src/experimental/Db.h:167]: I just looked at the pilot dialogue. You really dont need to sanatize the input. You *have* to do it now because you are looking for the line edits at the wrong place. You only need to change on line and there will be no "bad inputs"
```c++
// change this to
auto line_edits = parent()->findChildren<QLineEdit *>();
// this
auto line_edits = this->findChildren<QLineEdit*>();
```
pilotdialog

Once again the simplest solution is the best :D
Doesnt this clean up the db flow further? because now we can tweak the data in the encapsulated `Entry` and pass it in the `db::update(updated_entry)` and no sanatizing is necessary.
The dialog is constructed with the clicked `Entry` the user tweaks stuff we update the data in the `Entry` and pass it to `update`.

fiffty-50: Works for me, problem solved! I'll remove the cleanUp function entirely and merge then!

PR #14, opened 2020-12-08
Implementation of experimental::DB

fiffty-50:
Functions of experimental::DB are now implemented and open for discussion. Minor suggested adjustments to member names in Entry function, ref. https://github.com/fiffty-50/openpilotlog/wiki#code-style-conventions

### DIFF
--- a/src/experimental/Db.cpp
+++ b/src/experimental/Db.cpp
@@ -2,7 +2,7 @@
 
 using namespace experimental;
 
-bool DB::init()
+bool DB::connect()
 {
     const QString driver("QSQLITE");
 
@@ -31,4 +31,118 @@ bool DB::init()
         }
     }
     return true;
+}
+
+bool DB::commit(Entry entry)
+{
+    if (exists(entry)) {
+        return update(entry);
+    } else {
+        return insert(entry);
+    }
+}
+
+bool DB::remove(Entry entry)
+{
+    if (!exists(entry)) {
+        DEB("Error: Entry does not exist.");
+        return false;
+    }
+
+    QString statement = "DELETE FROM " + entry.position.first +
+            " WHERE _rowid_=" + QString::number(entry.position.second);
+    QSqlQuery q(statement);
+    //check result.
+    if (q.lastError().type() == QSqlError::NoError)
+    {
+        DEB("Entry " << entry.position.first << entry.position.second << " removed.");
+        return true;
+    } else {
+        DEB("Unable to delete.");
+        DEB("Query: " << statement);
+        DEB("Query Error: " << q.lastError().text());
+        return false;
+    }
+}
+
+bool DB::exists(Entry entry)
+{
+    if (entry.position.second == 0)
+        return false;
+
+    //Check database for row id
+    QString statement = "SELECT COUNT(*) FROM " + entry.position.first +
+            " WHERE _rowid_=" + QString::number(entry.position.second);
+    //this returns either 1 or 0 since row ids are unique
+    QSqlQuery q(statement);
+    q.next();
+    int rowId = q.value(0).toInt();
+    if (rowId) {
+        DEB("Entry exists with row ID: " << rowId);
+        return true;
+    } else {
+        DEB("Entry does not exist.");
+        return false;
+    }
+}
+
+
+bool DB::update(Entry updated_entry)
+{
+    auto data = updated_entry.getData();
+    QString statement = "UPDATE " + updated_entry.position.first + " SET ";
+    for (auto i = data.constBegin(); i != data.constEnd(); ++i) {
+        if (i.key() != QString()) {
+            statement += i.key() + QLatin1String("='") + i.value() + QLatin1String("', ");
+        } else {
+            DEB(i.key() << "is empty key. skipping.");
+        }
+    }
+    statement.chop(2); // Remove last comma
+    statement.append(QLatin1String(" WHERE _rowid_=") + QString::number(updated_entry.position.second));
+
+    DEB("UPDATE QUERY: " << statement);
+    QSqlQuery q(statement);
+    //check result.
+    if (q.lastError().type() == QSqlError::NoError)
+    {
+        DEB("Entry successfully committed.");
+        return true;
+    } else {
+        DEB("Unable to commit.");
+        DEB("Query: " << statement);
+        DEB("Query Error: " << q.lastError().text());
+        return false;
+    }
+}
+
+bool DB::insert(Entry newEntry)
+{
+    auto data = newEntry.getData();
+    DEB("Inserting...");
+    QString statement = "INSERT INTO " + newEntry.position.first + QLatin1String(" (");
+    QMap<QString, QString>::iterator i;
+    for (i = data.begin(); i != data.end(); ++i) {
+        statement += i.key() + QLatin1String(", ");
+    }
+    statement.chop(2);
+    statement += QLatin1String(") VALUES (");
+    for (i = data.begin(); i != data.end(); ++i) {
+        statement += QLatin1String("'") + i.value() + QLatin1String("', ");
+    }
+    statement.chop(2);
+    statement += QLatin1String(")");
+
+    QSqlQuery q(statement);
+    //check result.
+    if (q.lastError().type() == QSqlError::NoError)
+    {
+        DEB("Entry successfully committed.");
+        return true;
+    } else {
+        DEB("Unable to commit.");
+        DEB("Query: " << statement);
+        DEB("Query Error: " << q.lastError().text());
+        return false;
+    }
 }
--- a/src/experimental/Db.h
+++ b/src/experimental/Db.h
@@ -34,11 +34,12 @@ auto const DEFAULT_PILOT_POSITION = DataPosition("pilots", 0);
  */
 class Entry {
 public:
-    const DataPosition position;
+    DataPosition position;
 protected:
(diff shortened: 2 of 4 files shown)


src/gui/dialogues/newpilotdialog.cpp
```diff
@@ -149,7 +149,7 @@ void NewPilotDialog::submitForm()
     DEB("Creating Database Object...");
     QMap<QString, QString> newData;
 
-    auto line_edits = parent()->findChildren<QLineEdit *>();
+    auto line_edits = this->findChildren<QLineEdit *>();
 
     for (const auto &le : line_edits) {
```
KotzaBoss, line 154: `this->findChildren` Ensure that we have the correct line edits. The dialog *only* has the necessary information. Therefore the for loop can be simplyfied into what i had in the previous pull request #13.
```c++
for(auto& le : line_edits) {
    auto key = le->objectName().remove("LineEdit") // revove substring "LineEdit" and return new QString
    auto key = le->text();
    newData.insert(key, value);  // no need to check the key because it *will* exist and be valid
}
```

src/gui/dialogues/newpilotdialog.cpp
```diff
@@ -167,18 +167,20 @@ void NewPilotDialog::submitForm()
     newData.insert("displayname",displayName);
 
     using namespace experimental;
-    auto uin = newPilotInput(newData);
+    //auto uin = newPilotInput(newData);
+    auto entry = PilotEntry(newData);
+
+    if (role == Db::editExisting)
+        //entry.setPosition = oldEntry.position; //needs implementation in Entry class
```
KotzaBoss, line 174: If `editExisting` means that the user clicked an existing entry and poisition is pre-filled. No need for setPosition.

src/gui/dialogues/newpilotdialog.cpp
```diff
@@ -167,18 +167,20 @@ void NewPilotDialog::submitForm()
...
         /// if not i assume we should just emit inputRejected or smth?
-        if(!DB::insert(uin)) emit QDialog::rejected();
-        break;
-    case Db::editExisting:
-        DEB("updating entry with: " << newData);
-        if(!DB::update(uin)) emit QDialog::rejected();
-        break;
+        /// [F] We should do something, emitting reject might not be the best
+        /// option since it closes the dialog. Maybe showing a QMessageBox with
+        /// what has gone wrong?
```
KotzaBoss, line 182: Sure that is straight forward.
if `commit` returns false then **nothing** has been (should not be) altered and we simple re `submitForm` after prompting the error.

src/gui/dialogues/newpilotdialog.h
```diff
@@ -49,6 +49,8 @@ private slots:
 
     Pilot oldEntry;
 
+    QPair<QString, int> oldPosition;
+
```
KotzaBoss, line 53: this goes back to the pevious comments. Why would you have the position twice? one *in* the Entry and one *outside* of it? Leakage again i see xD

src/experimental/Db.h
```diff
@@ -51,112 +51,58 @@ class Entry {
...
-
-    if (data.isEmpty()) {
-        DEB("Object Contains no data. Aborting.");
-        return false;
-    }
-    QString statement = "INSERT INTO " + position.first + QLatin1String(" (");
-    for (auto i = data.cbegin(); i != data.cend(); ++i) {
-        statement += i.key() + QLatin1String(", ");
+    PilotEntry(TableData fromNewPilotDialog) : Entry::Entry(DataPosition("pilots", 0)) {
+        setData(fromNewPilotDialog);
```
KotzaBoss, line 57: This here is exactly what i mention about the position. If we are getting data from the dialog we *should* also have the position. It makes sense to do so since we populate the gui with entries from the database. We *know* where they are!

KotzaBoss (changes requested):
I like how its progressing. I see you prefer the subclassing of entry to functions. Well stick with that since you are comforable with it!
1. The primary thing is to be clear on what the `Entry` encapsulates and how it links tothe database.Specifically the position. Since the entry knows its position there is no need to keep it outside of it at the dialog level, whats the point. See comment for `Db.h` lines 56-57.
2. We must *ensure* that if any database operation returns false **nothing has been altered**. This is critical because whenever we get false as return value we are certain that our state hasnt changed and can simply ask the user to enter input again.

src/gui/dialogues/newpilotdialog.cpp
fiffty-50: thanks, I missed that. Included in new commit.

src/gui/dialogues/newpilotdialog.cpp
fiffty-50: adjusted.

src/gui/dialogues/newpilotdialog.cpp
fiffty-50: We could also handle a signal. DB could have for example these signals:
```cpp
...
signals:
    void commitSuccessful();
    void commitUnsuccessful(QString errorMsg);
...
```
then we could actually show what has gone wrong instead of just having a ```false``` return code.

KotzaBoss: Id like to study signals a bit more. I think lets keep the bool return for now because its very straight forward. If we use signals we propably have to handle them in the asyncronous way Qt handles them. We propably have to connect the Db signals with the dialog to provide a "reaction".

KotzaBoss:
```c++
class DB {
signals:
    void commitSuccessful();
    void commitUnsuccessful(QString errorMsg);
};

class NewPilotDialog {
slots:
    void onSuccessulCommit();
    void onUnsuccessfulCommit(QString msg);
    
    NewPilotDialog(...) {
        connect(&DB, &DB::commitSuccessful, this, &onSuccessfulCommit);
};
```
Im not sure if this will work because our database is static. Ill check it out once we have setup the dataflow. Its secondary atm. Just a different way of handling the same thing (success/unsuccess) on the db

src/gui/dialogues/newpilotdialog.h
fiffty-50: That was a workaround because I couldn't get the PilotEnty copied into the dialog with its copy operator deleted. How would you go about that, enable the copy operator for entry or is there a better way?

src/experimental/Db.h
fiffty-50: didn't see that comment, disregard my question in the comment above

src/gui/dialogues/newpilotdialog.cpp
KotzaBoss: I think i figured out how to do it but it may require the DB to be a singleton. It isnt toomuch of a hassle but it needs some work. Once we deal with this merge ill work on it.

src/gui/dialogues/newpilotdialog.h
KotzaBoss: You would simply pass in the existing entry upon construction and keep a pointer to it. you only need the `oldEntry` which has position encapsulated already

src/experimental/Db.cpp
```diff
@@ -32,3 +32,117 @@ bool DB::init()
...
+        DEB("Unable to delete.");
+        DEB("Query: " << statement);
+        DEB("Query Error: " << q.lastError().text());
+        return false;
+    }
+}
+
+bool DB::exists(Entry entry)
+{
+    if (entry.position.second == 0)
```
KotzaBoss, line 70: If you dont like this we can make position a custom type with better names for attributes.
```c++    
struct DP : DataPosition {
    TableName table;
    RowId row_id;
    DP(TableName t, RowId i) : DataPosition(t, i) {}
};
```
now we can write `.table` or `.row_id` for extra clarity for the future

src/experimental/Db.h
```diff
@@ -51,112 +51,58 @@ class Entry {
...
 /*!
  * \brief The DB class encapsulates the SQL database by providing fast access
  * to hot database data.
  */
 class DB {
 private:
-    static QStringList tableNames;
-    static QMap<TableName, QStringList> tableColumns;
+    static inline QStringList                   tableNames;
+    static inline QMap<TableName, QStringList>  tableColumns;
```
KotzaBoss, line 70: what does `inline` provide here?

KotzaBoss (approved):
Seems good for now. I want to do some research on default copy constructors because there are *alot* of semantics involved. 
Also because of the copy constructor we may not need the edit roles in the first place since we call commit which does the check for us!

src/experimental/Db.h
fiffty-50: Wouldn't compile without it, the problem ocurred when connect() tried writing to these variables. I found [this](url=https://stackoverflow.com/questions/9656941/why-cant-i-initialize-non-const-static-member-or-static-array-in-class) explication and went with it.



> It's because there can only be one definition of A::a that all the translation units use.

>If you performed static int a = 3; in a class in a header included in all a translation units then you'd get multiple definitions. Therefore, non out-of-line definition of a static is forcibly made a compiler error.

>Using static inline or static const remedies this. static inline only concretises the symbol if it is used in the translation unit and ensures the linker only selects and leaves one copy if it's defined in multiple translation units due to it being in a comdat group. const at file scope makes the compiler never emit a symbol because it's always substituted immediately in the code unless extern is used, which is not permitted in a class.

PR #18, opened 2020-12-10
Devel db get entry data

fiffty-50:
DataBase has the following new functions:
database - return the database currently in use
disconnect - disconnect the database
getEntryData(Position)
getEntry(DataPosition)
getPilotEntry(RowId)

The NewPilot Dialog is refactored to make use of the new experimental classes. Editing pilots and adding new pilots seems to work.

### DIFF
--- a/src/experimental/DataBase.cpp
+++ b/src/experimental/DataBase.cpp
@@ -2,6 +2,11 @@
 
 namespace experimental {
 
+/// [F] suggestions for signals:
+///
+/// commitSuccessful()
+/// commitUnSuccessful(QString sqlError, QString statement)
+
 DataBase* DataBase::instance = nullptr;
 
 DataBase* DataBase::getInstance()
@@ -130,11 +135,13 @@ bool DataBase::update(Entry updated_entry)
     if (q.lastError().type() == QSqlError::NoError)
     {
         DEB("Entry successfully committed.");
+        /// [F] emit commitSuccessful();
         return true;
     } else {
         DEB("Unable to commit.");
         DEB("Query: " << statement);
         DEB("Query Error: " << q.lastError().text());
+        /// [F] emit commitUnsuccessful(q.lastError().type(), statement);
         return false;
     }
 }
@@ -161,11 +168,13 @@ bool DataBase::insert(Entry newEntry)
     if (q.lastError().type() == QSqlError::NoError)
     {
         DEB("Entry successfully committed.");
+        /// [F] emit commitSuccessful();
         return true;
     } else {
         DEB("Unable to commit.");
         DEB("Query: " << statement);
         DEB("Query Error: " << q.lastError().text());
+        /// [F] emit commitUnsuccessful(q.lastError().type(), statement);
         return false;
     }
 
@@ -176,7 +185,6 @@ TableData DataBase::getEntryData(DataPosition dataPosition)
     // check table exists
     if (!tableNames.contains(dataPosition.first)) {
         DEB(dataPosition.first << " not a table in the database. Unable to retreive Entry data.");
-        // emit databaseError
         return TableData();
     }
 
@@ -188,14 +196,12 @@ TableData DataBase::getEntryData(DataPosition dataPosition)
     if (checkQuery.lastError().type() != QSqlError::NoError) {
         DEB("SQL error: " << checkQuery.lastError().text());
         DEB("Statement: " << statement);
-        // emit sqlError(selectQuery.lastError().text(), statement)
         return TableData();
     }
 
     checkQuery.next();
     if (checkQuery.value(0).toInt() == 0) {
         DEB("No Entry found for row id: " << dataPosition.second );
-        // emit databaseError("No Entry found for row id: " + dataPosition.second)
         return TableData();
     }
 
@@ -208,7 +214,6 @@ TableData DataBase::getEntryData(DataPosition dataPosition)
     if (selectQuery.lastError().type() != QSqlError::NoError) {
         DEB("SQL error: " << selectQuery.lastError().text());
         DEB("Statement: " << statement);
-        // emit sqlError(selectQuery.lastError().text(), statement)
         return TableData();
     }
 
--- a/src/experimental/DataBase.cpp
+++ b/src/experimental/DataBase.cpp
@@ -2,11 +2,12 @@
 
 namespace experimental {
 
-DataBase* const DataBase::instance = nullptr;
+DataBase* DataBase::instance = nullptr;
 
 DataBase* DataBase::getInstance()
 {
-    return instance ?: new DataBase();
+    if (!instance) instance = new DataBase();
+    return instance;
 }
 
 bool DataBase::connect()
@@ -37,6 +38,7 @@ bool DataBase::connect()
             tableColumns.insert(table, columnNames);
         }
     }
+    DEB("Database Tables: " << tableNames);
     return true;
 }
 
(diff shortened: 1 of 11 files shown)


src/experimental/DataBase.cpp
```diff
@@ -236,6 +251,20 @@ TableData DataBase::getEntryDataNew(DataPosition dataPosition)
         entryData.insert(column, model.record(0).value(column).toString());
     }
     return entryData;
+}*/
+
+Entry DataBase::getEntry(DataPosition dataPosition)
+{
+    Entry entry(dataPosition);
```
KotzaBoss, line 258: Can i take this as a cue to stick to copy constructor initialisation? (insted of operator= copy)

src/experimental/DataBase.cpp
```diff
@@ -42,6 +42,20 @@ bool DataBase::connect()
...
+    auto db = DataBase::database();
+    db.close();
+    db.removeDatabase(db.connectionName());
+    DEB("Database connection closed.");
+}
+
+QSqlDatabase DataBase::database()
+{
+    auto db = QSqlDatabase::database("qt_sql_default_connection");
+    return db;
```
KotzaBoss, line 56: No need to assign and then return
```c++
return QSqlDataBase::database("...");
```

src/experimental/DataBase.h
```diff
@@ -43,7 +38,13 @@ class DataBase {
...
      * \brief closes the database connection.
      */
-    bool disconnect();
+    void disconnect();
+
+    /*!
+     * \brief Can be used to access the database connection.
+     * \return The QSqlDatabase object pertaining to the connection.
+     */
+    QSqlDatabase database();
```
KotzaBoss, line 47: you call this fucntion statically at Database::disconnect. since it returns something handled by Qt shouldnt it be static?

src/experimental/DataBase.h
```diff
@@ -76,7 +77,22 @@ class DataBase {
...
+    /*!
+     * \brief retreives a PilotEntry from the database.
+     *
+     * This function is a wrapper for DataBase::getEntry(DataPosition),
+     * where the table is already set and which returns a PilotEntry
+     * instead of an Entry. It allows for easy access to a pilot entry
+     * with only the RowId required as input.
+     */
+    PilotEntry getPilotEntry(RowId);
+    ///[F] the same can easily be implemented for tails/flights
```
KotzaBoss, line 94: Im sure we will see code duplication here. But lets write it first before cleaning it up so that we are sure we are on the right path.

src/experimental/Entry.cpp
```diff
@@ -4,7 +4,21 @@ namespace experimental {
 
 Entry::Entry(DataPosition position_)
     : position(position_)
-{}
+{
+//    position = position_;
+}
+/// [F] from what I understand the initialiser list should do exactly the same as
```
KotzaBoss, line 10: What you gain with the initlizers is performance. Especially for our simple classes there really isnt any reason to put code in the braces because we only initilise data. One interesting use for initilizers you can find in the `DataPosition` class. Can you understand what happens? and why i do it that way?

src/experimental/Entry.cpp
```diff
@@ -17,8 +31,9 @@ const TableData& Entry::getData()
 }
 
 PilotEntry::PilotEntry(int row_id)
-    : Entry::Entry(DataPosition("pilots", row_id))
-{}
+{
+    position = DataPosition("pilots", row_id);
+}
```
KotzaBoss, line 36: see above comment.
Keep previous.

src/gui/dialogues/newpilotdialog.h
```diff
@@ -38,19 +38,15 @@ class NewPilotDialog : public QDialog
 {
     Q_OBJECT
 public:
-    explicit NewPilotDialog(Db::editRole, QWidget *parent = nullptr);
-    explicit NewPilotDialog(Pilot, Db::editRole, QWidget *parent = nullptr);
-    explicit NewPilotDialog(experimental::PilotEntry oldEntry, Db::editRole, QWidget *parent = nullptr);
+    explicit NewPilotDialog(QWidget *parent = nullptr);
+    explicit NewPilotDialog(int rowId, QWidget *parent = nullptr);
```
KotzaBoss, line 42: aaaah yeshhh thats what i like to see. Simple cleaner constructors. Almost better than ice cream... almost...

src/gui/widgets/homewidget.cpp
```diff
@@ -40,9 +40,17 @@ void HomeWidget::on_pushButton_clicked()
...
+    auto entry = DB()->getPilotEntry(7);
+    DEB(entry.position);
+    DEB(entry.getData());
+
+    PilotEntry copiedEntry;
+    copiedEntry = entry;
+    DEB(copiedEntry.position);
+    DEB(copiedEntry.getData());
+
+    /*long customFunc = 0;
```
KotzaBoss, line 53: Did you perhaps forget to cleanup debugging?

src/gui/widgets/homewidget.cpp
```diff
@@ -59,5 +67,5 @@ void HomeWidget::on_pushButton_clicked()
         qSqlTableModelFunc += duration2.count();
     }
     DEB("Average execution time: (custom Func)        " << customFunc/10000 << "ms");
-    DEB("Average execution time: (qSqlTableModelFunc) " << qSqlTableModelFunc/10000 << "ms");
+    DEB("Average execution time: (qSqlTableModelFunc) " << qSqlTableModelFunc/10000 << "ms");*/
```
KotzaBoss: so you werent joking our *is* faster.

src/gui/widgets/pilotswidget.cpp
```diff
@@ -39,12 +39,12 @@ void PilotsWidget::tableView_selectionChanged()//const QItemSelection &index, co
     selectedPilots.clear();
 
     for (const auto& row : selection->selectedRows()) {
-        selectedPilots << row.data().toInt();
+        selectedPilots.append(row.data().toInt());
```
KotzaBoss, line 42: still a bit unclear on this. When selection changes what does `data()` return?

KotzaBoss (changes requested):
Looking good. Code is cleaning up. We should ensure PilotDialog works well before branching of to the other dialogs. 
Keep the initializers for construction. And consider making `DataBase::database()` static (see comment about it)

src/gui/widgets/homewidget.cpp
fiffty-50: indeed xD

src/gui/dialogues/newpilotdialog.cpp
```diff
@@ -65,8 +65,12 @@ NewPilotDialog::NewPilotDialog(QWidget *parent) :
     setupCompleter();
 
     using namespace experimental;
-    //    connect(DB(), &DataBase::commitUnsuccessful,
-    //            this, &NewPilotDialog::onCommitUnsuccessful);
+
+    QObject::connect(DB(), &DataBase::commitSuccessful,
+            this, &NewPilotDialog::onCommitSuccessful);
+    QObject::connect(DB(), &DataBase::commitUnsuccessful,
+            this, &NewPilotDialog::onCommitUnsuccessful);
```
KotzaBoss, line 72: Maybe do the connection in a higher level. Instead of the dialog re-connecting itself. Have the `mainwindow` connect `on_newPilot_triggered`?

KotzaBoss (changes requested):
Move connection from dialog to mainwindow
```c++
void MainWindow::on_actionNewPilot_triggered()
{
    NewPilotDialog np =NewPilotDialog(Db::createNew, this);
    QObject::connect(DB(), &DataBase::commitSuccessFull, &np, &NewPilotDialog::onsuccess);
     np.exec();
}

PR #35, opened 2020-12-25
Testing pull request creation from gitkraken with some minor commits

fiffty-50:
Testing pull request creation from gitkraken with some minor commits


src/gui/dialogues/firstrundialog.cpp
```diff
@@ -1,6 +1,10 @@
 #include "firstrundialog.h"
 #include "ui_firstrundialog.h"
 
+const auto TEMPLATE_URL = QLatin1String("https://raw.githubusercontent.com/fiffty-50/openpilotlog/develop/assets/database/templates/");
```
KotzaBoss, line 4: Why do you use QLatin1String everywhere and not just QString?

fiffty-50: to improve [performance](https://doc.qt.io/qt-5/qlatin1string.html#details) - I am also trying to use [QStringLiteral](https://doc.qt.io/qt-5/qstring.html#QStringLiteral) where possible for the same reason.

PR #38, opened 2020-12-29
Develop removing obsolete code

fiffty-50:
Namespace experimental removed, moved files into regular source structure and adjusted includes etc.


mainwindow.cpp
```diff
@@ -76,7 +77,7 @@ MainWindow::MainWindow(QWidget *parent)
     /// [F] I understand how it is annoying to not have the database
     /// working when something has changed. Hopefully this check
     /// helps to avoid that in the future!
-    const int DATABASE_REVISION_NUMBER = 14;
+    const int DATABASE_REVISION_NUMBER = 15;
```
KotzaBoss, line 80: Instead of making a const variable you should consider doing the following.
```c++
// ADatabase.h
#define DATABASE_VERSION 15
```
That way we can use the this imformatino in a very clean way:
```c++
// any_file_that_interfaces_with_db.h

void function(){
#if DATABASE < 15
    // code for database <15
#else
    // some cool optimisation that is not supported by <15
#endif
}
```
That way the code in the true branch of the preprocessor #if will be put in the file.
And also during developement should there be any hiccup between database revisions we can alleviate it with a preprocessor making it easier than going through the "what is the database version, let me pull it etc etc"

src/database/declarations.h
```diff
@@ -1,17 +1,56 @@
...
+using ColumnData = QPair<ColName, ColData>;
+using ColumnNames = QStringList;
+using TableColumns = QMap<TableName, ColumnNames>;
+
+// [G]: Needs some work. Inheriting from QPair may be helpful but
+// may also be overkill. Lets determine the specific uses of DataPosition
+// and provide our own interface i would say.
+// [F]: Good idea! Implementing something similar to first and second methods
+// of QPair would be useful to carry over, or some other way of quickly and
+// unambiguously accessing the elements.
```
KotzaBoss, line 32: first and second arent methods. They are members. QPair is a struct so everything is public unless specified

src/database/declarations.h
```diff
@@ -1,17 +1,56 @@
...
-// [F]: These constants deviate slightly from const naming convention to reflect their sql column name.
+/*!
+ * \brief An alias for QString
+ *
+ * Very long description *with* **markdown?**
+ * - - -
+ * # Header
+ */
+using ColName = QString;
+using ColData = QVariant;
```
KotzaBoss, line 14: I dont see any reason to complicate this. Everything from the database is a string anyway. Because you have it as a variant you make the code in the entries needlessly complex by "casting" it to `toString` all the time.

src/gui/dialogues/newflightdialog.cpp
```diff
@@ -481,9 +485,9 @@ void NewFlightDialog::fillDeductibleData()
  * no input validation is done in this step and input data is assumed to be valid.
  * \return
  */
-TableData NewFlightDialog::collectInput()
+RowData NewFlightDialog::collectInput()
 {
-    TableData newData;
+    RowData newData;
     DEB("Collecting Input...");
```
KotzaBoss, line 491: throwaway comment.
We can define DEB as
```c++
#define DEB qDebug() << ___PRETTY_FUNCTION___
DEB << "..." << "asdf";
```
we just shorten the akward typing of qDebug() to DEB no need for parentheses

src/classes/aflightentry.h
```diff
@@ -1,18 +1,33 @@
...
 class AFlightEntry : public AEntry {
 public:
     AFlightEntry();
     AFlightEntry(const AFlightEntry& pe) = default;
     AFlightEntry& operator=(const AFlightEntry& pe) = default;
     AFlightEntry(int row_id);
-    AFlightEntry(TableData table_data);
+    AFlightEntry(RowData table_data);
 
     enum pilot { pic, sic, thirdPilot };
```
KotzaBoss, line 32: perhaps rename to `pilotPosition` or some aviation specific name for what the pilot is.
Just pilot feels a bit ambiguous. Especially for someone not in the industry!

src/classes/aflightentry.cpp
```diff
@@ -0,0 +1,73 @@
...
+        return foreign_pilot.name();
+    }
+    case pilot::thirdPilot: {
+        auto foreign_pilot = aDB()->resolveForeignPilot(tableData.value(DB_FLIGHTS_THIRDPILOT).toInt());
+        return foreign_pilot.name();
+        break;
+    } // case scope
+    } // switch (pilot_)
+    return QString();
+}
```
KotzaBoss, line 73: This is a bit suspect use of the enum.
`enum something {...}` is a plain enum. Even though you can write `something::A` you are misusing the scope operator. Enum classes are prefered over enums in general because they *demand* the use of the `::` operator and are more type safe. Because they `pilots` enum is already encapsulated the risk isnt too great but at least show that its plain in its use.
```c++
switch(pilot_) {
    case AFlightEntry::pic:  // explicit
    case sic: // implicit since you are working in the method scope (in this context this is ok)
```
If it was enum class...
```c++
switch(pilot_) {
    case AFlightEntry::pilots::pic:  // explicit root
    case pilots::sic:  // implicit since in method
    case sic:  // wont compile
```

src/classes/atailentry.cpp
```diff
@@ -17,36 +17,32 @@
...
-    if (!getData().value(QLatin1String("model")).toString().isEmpty())
-        type_string.append(getData().value(QLatin1String("model")).toString());
-    if (!getData().value(QLatin1String("variant")).toString().isEmpty())
-        type_string.append('-' + getData().value(QLatin1String("variant")).toString() + ' ');
+    if (!getData().value(DB_TAILS_MAKE).toString().isEmpty())
+        type_string.append(getData().value(DB_TAILS_MAKE).toString() + ' ');
+    if (!getData().value(DB_TAILS_MODEL).toString().isEmpty())
+        type_string.append(getData().value(DB_TAILS_MODEL).toString());
+    if (!getData().value(DB_TAILS_VARIANT).toString().isEmpty())
+        type_string.append('-' + getData().value(DB_TAILS_VARIANT).toString() + ' ');
```
KotzaBoss, line 45: This doesnt seem like clean code.
Because you are in the scope of the Tail entry there is no need to call the getter, the class handles its private data without a getter. 
Also see my other comment about RowData being a variant. 
Tl;dr RowData should be QString since everything read is a string anyway. If a class needs to convert a "time string" into ISO class thats its problem. The basic data type from the DB is QString not QVariant. Unless im missing something

KotzaBoss (changes requested):
Check my comments for specifics. In short i suggest:
### General
- `RowData` from `QVariant` to `QString`
-  Database version from variable to preprocessor macro to enable use of directives for backwards compatibility if necessary
- Change name of `AFlightEntry` enum from `pilot` to something more explicit
### Subjective
- Change the DEB macro to something with less characters required to type

src/database/declarations.h
fiffty-50: On the contrary, having this element as a QVariant actually avoids quite a lot of type casting. Most of the fields in the database are actually integers, not strings. The database holds different types, integers, text, numerics etc. If we cast all those into QString, we have to convert them back to their actual type before using them, effectively casting for example int to qstring to int.

There are also benefits when entering data into the database. When using QVariant, we can directly enter integers or strings depending on the database type into the map. Then, when creating the query to submit it, QVariant automatically binds the correct type to the insert query.

For example:

```cpp
// NewTailDialog
new_data.insert(DB_TAILS_MULTIPILOT, ui->operationComboBox->currentIndex() - 1); 
// here we can directly insert the integer instead of doing:
new_data.insert(QString::number(DB_TAILS_MULTIPILOT, ui->operationComboBox->currentIndex() - 1)); 
// here we would cast an integer to string, only to bind it to a string when we are creating a query, send a string to the database, and convert it to an integer that's stored in the database again.

// more examples
// NewFlightDialog
newData.insert(DB_FLIGHTS_TBLK, block_minutes);
newData.insert(DB_FLIGHTS_ACFT, tailsIdMap.value(ui->acftLineEdit->text()));

// ADatabase.cpp
bool ADatabase::insert(AEntry new_entry) {
query.prepare("statement with ? placeholders");
query.addbindvalue; 
// binds int to int and string to string instead of sending all strings to the database, which then has to do the casting
query.exec();
}

ADatabase::getEntryData(DataPosition data_position) {
// select data from database
//...
RowData entry_data;

    for (const auto &column : tableColumns.value(data_position.first)) {
        entry_data.insert(column, select_query.value(column));
    }
// QSqlQuery returns a QVariant already, so creating the map like this is much faster than doing this
RowData entry_data;

    for (const auto &column : tableColumns.value(data_position.first)) {
        entry_data.insert(column, select_query.value(column).toString);
    }
}
```

The same is true the other way around, when retreiving data from the database. Retreiving integers only to cast them to strings, and then do `string.toInt()` again seems rather wasteful and unneccessary. A `QMap<QString, QVariant>` is thus more efficient and also a more accurate representation of a database entry in my opinion.

mainwindow.cpp
fiffty-50: Good point, I'll implement it!

src/classes/atailentry.cpp
fiffty-50: removed the getter and accessed data directly, I guess that was a leftover from when tableData was protected.

See comment above for QVariant vs QString

src/database/adatabase.cpp
```diff
@@ -18,6 +18,7 @@
 #include "adatabase.h"
 #include "src/testing/adebug.h"
 
+#define DATABASE_VERSION 15
```
KotzaBoss, line 21: it should be in the .h file. The point is that its going to be #included to get the value. Right now its... not good

src/classes/atailentry.cpp
KotzaBoss: even if it was private the class can access any data it encapsulates.

src/database/declarations.h
```diff
@@ -40,6 +40,21 @@ struct DataPosition : QPair<TableName, RowId> {
...
+struct DataPosition {
+    TableName tableName;
+    RowId rowId;
+    DataPosition()
+        : tableName(TableName())
+    {};
+    DataPosition(TableName table_name, RowId row_id)
+        : tableName(table_name), rowId(row_id)
+    {};
+
```
KotzaBoss, line 57: The main problem is the debug integration with Qt. otherwise its what we want atm.

fiffty-50: fixed that

PR #40, opened 2021-01-01
Pulling into primary rework branch

KotzaBoss:
Major changes to FirstRunDialog.
All hardcoded ini keys encapsulated in ASettings.
Review needed:
- Are all keys encapsulated?
- Is naming consistent? (its not)

After ensuring that the logic is sound we move to debugging.


fiffty-50:
Merging now for debugging, we will address your comment before merging into develop

PR #41, opened 2021-01-01
ASettings mass refactor

KotzaBoss:
ASettings has possible keys of config file as enums.

Require debugging of ADataBaseSetup. 
Focus on FirstRunDialog, ADataBaseSetup and ASettings


src/gui/dialogues/firstrundialog.cpp
```diff
@@ -33,149 +36,144 @@ FirstRunDialog::~FirstRunDialog()
...
+    data.insert(ASettings::stringOfKey(ASettings::UserData::EmployeeID), ui->employeeidLineEdit->text());
+    data.insert(ASettings::stringOfKey(ASettings::UserData::Phone), ui->phoneLineEdit->text());
+    data.insert(ASettings::stringOfKey(ASettings::UserData::Email), ui->emailLineEdit->text());
+
+    auto db_fail_msg_box = QMessageBox(QMessageBox::Critical, QStringLiteral("Database setup failed"),
+                                       QStringLiteral("Errors have ocurred creating the database."
+                                                      "Without a working database The application will not be usable."));
+    // [G]: Im abit confused on the logic here
+    // why do you write setup complete twice?
+    if (!setupDatabase()) {
```
KotzaBoss, line 117: This returns false. Tracing back the problem lead to error in ADataBaseSetup::createSchemata look for comment on that function

src/classes/asettings.cpp
```diff
@@ -19,23 +19,142 @@
...
+QVariant ASettings::read(const UserData key)
+{ return QSettings().value(pathOfKey(key)); }
+
+void ASettings::write(const UserData key, const QVariant &val)
+{ QSettings().setValue(pathOfKey(key), val); }
+
+//
+// QString conversion PATH
+//
+QString ASettings::pathOfKey (const ASettings::FlightLogging key)
```
fiffty-50, line 118: `groupOfKey`, `sectionOfKey` or something similar  would be a more suitable name. path makes it seem like a directory or file system related value, whereas it is actually just the QSettings class' way of logically [grouping together items](https://doc.qt.io/qt-5/qsettings.html#section-and-key-syntax) of the same category to read and write them under the same header in the INI file.

src/classes/asettings.h
```diff
@@ -24,20 +24,180 @@
...
+    static QVariant read(const FlightLogging key);
+    static void write(const UserData key, const QVariant &val);
+
+    static QVariant read(const UserData key);
+    static void write(const FlightLogging key, const QVariant &val);
+
+    static QVariant read(const NewFlight key);
+    static void write(const NewFlight key, const QVariant &val);
+
+    // [G]: enum class may be making it abit too strict perhaps?
```
fiffty-50, line 98: I prefer the enum class approach as you laid out above. It is more accessible (human readable) and easier to interact with. Using a string-based approach is also closer to the settings file itself. After all, it's just a plain text file and since the program is still quite volatile as you correctly pointed out, I think it is better to work with a more accessible way of accessing the settings - Keep it simple stupid and let's consider revisiting this topic at a later stage!

fiffty-50 (changes requested):
Thank you for these commits!

Database creation and setup seems to work on my end so let's troubleshoot on discord and figure out the exact problem. I have a theory - there are a few quirks around creating the database for the first time and the way we save the database layout as member variables in the database class. It might be better to access them dynamically where needed instead of writing them once when calling ADatabase::connect().

For the settings, let's stick with using the enum class for now, I have laid out the reasoning in a comment there, in short, [KISS](https://en.wikipedia.org/wiki/KISS_principle)!

Nice work on the standardpaths, looks much better than before!

src/classes/asettings.h
```diff
@@ -24,20 +24,180 @@
...
+    enum class UserData {
+        LastName,
+        FirstName,
+        Company,
+        EmployeeID,
+        Phone,
+        Email,
+        DisplaySelfAs,
+        Alias,
+        AcSortColumn, // not sure what ac is
```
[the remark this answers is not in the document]
KotzaBoss: Ac might as well be Assassins Creed.
Acft should be the minimum for Aircraft

src/classes/asettings.cpp
```diff
@@ -19,23 +19,142 @@
 #include "src/astandardpaths.h"
 #include <QSettings>
 
+
+QMap<ASettings::Main, QString> ASettings::mainMap = {
+    {Main::Theme,   QStringLiteral("theme")},
+    {Main::ThemeID, QStringLiteral("theme_id")}, // inconsistent naming
```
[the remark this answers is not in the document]
KotzaBoss: I got confused with the parameter to the function being named theme_id.
Perhaps they should be the same.

Will remove it.

KotzaBoss:
Need to go through all(?) dialogs and make sure that if alt-f4 is pressed we control the outcome. Even when error messages that should stop the program get alt-f4-ed the program wrongly continues.

See [this](https://stackoverflow.com/questions/10369826/how-can-i-disable-alt-f4-window-closing-using-qt) for potential solution.

KotzaBoss:
> Thank you for these commits!
> 
> Database creation and setup seems to work on my end so let's troubleshoot on discord and figure out the exact problem. I have a theory - there are a few quirks around creating the database for the first time and the way we save the database layout as member variables in the database class. It might be better to access them dynamically where needed instead of writing them once when calling ADatabase::connect().
> 
> For the settings, let's stick with using the enum class for now, I have laid out the reasoning in a comment there, in short, [KISS](https://en.wikipedia.org/wiki/KISS_principle)!
> 
> Nice work on the standardpaths, looks much better than before!

Not sure about what dynamically means. Can you elaborate on what part of the process we should fix up?

PR #42, opened 2021-01-03
Application style encapsulation in AStyle

KotzaBoss:
Consolidated the style management logic in class `AStyle`. Stylesheet selection removed because Qt offers both light and dark styles. Should an interesting stylesheet be wanted further extension should be added.

### Caveats
Changing styles works however after 1 or 2 changes results are inconsistent and seem to blend into one another. Also avoid bb themes because they crash the program. They arent good anyway so may be removed in the future.


main.cpp
```diff
@@ -20,8 +20,9 @@
 #include "src/classes/arunguard.h"
 #include "src/database/adatabase.h"
 #include "src/classes/asettings.h"
-#include "src/astandardpaths.h"
+#include "src/classes/astandardpaths.h"
 #include "src/classes/asettings.h"
+#include "src/astyle.h"
```
KotzaBoss, line 25: Forgot to put in proper dir. Will do before merging

src/gui/widgets/logbookwidget.cpp
```diff
@@ -54,6 +54,7 @@ LogbookWidget::LogbookWidget(QWidget *parent) :
 
 LogbookWidget::~LogbookWidget()
 {
+    delete displayModel;  // [G]: previously leaked from logbookwidget.cpp:89
```
fiffty-50, line 57: wouldn't it be easier to create the model with this as parent and let qt handle this?
```
89: displayModel = new QSqlTableModel(this);
```

src/database/adatabasesetup.cpp
```diff
@@ -299,16 +299,18 @@ bool ADataBaseSetup::downloadTemplates()
...
-            return false;
-        }
-        DEB << "Backing up old database as: " << backup_name;
+    if(!database_file.exists()) {
+        DEB << "No Database to backup, returning.";
+        return true;
+    }
+    auto date_string = QDateTime::currentDateTime().toString(Qt::ISODate);
+    auto backup_dir = QDir(AStandardPaths::pathTo(AStandardPaths::DatabaseBackup));
+    auto backup_name = database_file.baseName() + "-backup-" + date_string + ".bak";
```
fiffty-50, line 308: maybe keep the extension as .db since the backup is already in the file name and it makes it clearer what kind of file we are dealing with?

src/classes/astyle.cpp
```diff
@@ -5,8 +5,13 @@
 #include "src/testing/adebug.h"
 #include "src/classes/asettings.h"
 
+#ifdef __linux__
```
fiffty-50, line 8: We should look into how fusion vs native looks on windows, maybe fusion might be a suitable default on all platforms?

src/classes/astandardpaths.cpp
```diff
@@ -1,47 +1,47 @@
...
 
 void AStandardPaths::setup()
 {
     auto settings_location = QStandardPaths::AppConfigLocation;
     auto data_location = QStandardPaths::AppDataLocation;
-    paths = {  // [G]: potential rename to `dirs`
+    dirs = {  // [G]: potential rename to `dirs`
         {Database, QStandardPaths::writableLocation(data_location)},
         {Templates, QDir(QStandardPaths::writableLocation(data_location)).filePath("templates")},
         {Settings, QStandardPaths::writableLocation(settings_location)},
```
fiffty-50, line 13: I think this is obsolete since we are using the default constructor of QSettings now. This results in the ini file being created in 
`~/.config/opl/openPilotLog.ini`, so creating an empty folder `openPilotLog` in the same directory where `openPilotLog.ini` is located looks dirty.
tree

fiffty-50 (commented):
Nice work!

Once we fixed the style/stylesheet selectors we are good for merge!

src/gui/widgets/logbookwidget.cpp
KotzaBoss: fixed

src/database/adatabasesetup.cpp
KotzaBoss: actually i didnt even look at the fact that backup is there.
A suggestion for naming would be
`logbook.db -> logbook-ISODATE.db.bak`
that way have both data and backup in a clean fairly conventional way

src/classes/astandardpaths.cpp
KotzaBoss: So the default constructor doesnt conform to AppConfigLocation?

src/database/adatabasesetup.cpp
fiffty-50: That's good for linux but windows is quite picky with files that have more than one `.` in them due to the whole file extension handling. Maybe `logbook.db -> logbook_bak_isodate.db`?

src/classes/astandardpaths.cpp
fiffty-50: it does use standard [locations](https://doc.qt.io/qt-5/qsettings.html#locations-where-application-settings-are-stored), depending on the [scope](https://doc.qt.io/qt-5/qsettings.html#setPath) but the path does only include the orgname, the appname is used for the settings file itself:
 `$HOME/.config/opl/openPilotLog.ini`
The path is automatically created if needed so we don't need to manually handle it.

src/database/adatabasesetup.cpp
KotzaBoss: If i knew this isnt a public log i would say lots of stuff... -,-
yours seems safest. Changing

src/classes/astandardpaths.cpp
KotzaBoss: so we are ok QSettings handles its as it should even if it seems off in our eyes.
Any custom configs will go to opl/openPilotLog wouldnt it?

fiffty-50: yes, `opl/openPilotLog.ini`

PR #43, opened 2021-01-04
Develop debugwidget refactor

fiffty-50:
DebugWidget now makes use of standardpaths, no more need to restart application after database changes.

Update of ADatabase to address an issue that ocurred when the database was modified, but the tableNames and columnNames member variables had outdated information since they were only once written when connect() was called. See commit [159c391](https://github.com/fiffty-50/openpilotlog/commit/159c391d9b75dd58608efe7a7f2514d546ce2c14) for a detailed description of the issue.

### DIFF
--- a/src/database/adatabase.cpp
+++ b/src/database/adatabase.cpp
@@ -33,33 +33,41 @@ QString ADatabaseError::text() const
 
 ADatabase* ADatabase::instance = nullptr;
 
-//[F]: This is what I mentioned the other day, instead of writing the table names
-// to a member variable once at connect(), we are accessing the layout at the moment we query it.
-// While being a little more expensive, this makes the app much more robust because
-// it always returns an accurate description of the database, regardless of whether
-// it is already created or not. Also it encapsulates this functionality here instead
-// of being included in connect()
 /*!
- * \brief Return the names of the tables in the database.
+ * \brief Return the names of a given table in the database.
  */
-TableNames ADatabase::getTableNames() const
+ColumnNames ADatabase::getTableColumns(TableName table_name) const
 {
-    auto db = ADatabase::database();
-    return db.tables();
+    return tableColumns.value(table_name);
 }
 
 /*!
- * \brief Return the names of the columns for a table in the database
+ * \brief Return the names of all tables in the database
  */
-TableColumns ADatabase::getTableColumns(TableName table_name) const
+TableNames ADatabase::getTableNames() const
+{
+    return tableNames;
+}
+
+/*!
+ * \brief Updates the member variables tableNames and tableColumns with up-to-date layout information
+ * if the database has been altered. This function is normally only required during database setup or maintenance.
+ */
+void ADatabase::updateLayout()
 {
     auto db = ADatabase::database();
-    TableColumns table_columns;
-    QSqlRecord fields = db.record(table_name);
-    for (int i = 0; i < fields.count(); i++) {
-        table_columns.append(fields.field(i).name());
+    tableNames = db.tables();
+
+    tableColumns.clear();
+    for (const auto &table_name : tableNames) {
+        ColumnNames table_columns;
+        QSqlRecord fields = db.record(table_name);
+        for (int i = 0; i < fields.count(); i++) {
+            table_columns.append(fields.field(i).name());
+        }
+        tableColumns.insert(table_name, table_columns);
     }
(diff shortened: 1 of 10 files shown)


src/database/adatabase.cpp
```diff
@@ -33,14 +33,33 @@ QString ADatabaseError::text() const
...
+// of being included in connect()
+/*!
+ * \brief Return the names of the tables in the database.
+ */
 TableNames ADatabase::getTableNames() const
 {
-    return tableNames;
+    auto db = ADatabase::database();
+    return db.tables();
 }
```
KotzaBoss, line 49: While this needs benchmarking to determine if its better. 
At a glance i see it as worse. I dont know how db.tables() works in the background but re-reading the tables with each query doesnt seem efficient.
reading them upfront and simply updating them seems more reasonable. 

Once the program starts the database is read, the local data gets populated for quicker access.
Any updates to the db update the local data and at the end the sql database is edited.

Mechanically an accessor can remove the clunky tableData.value(...)
I assume we will not see any preformance difference right now, so we should consider setting up a testing module.
Im up for it if its not particularly interesting for you.

src/gui/dialogues/firstrundialog.cpp
```diff
@@ -120,8 +120,6 @@ bool FirstRunDialog::finish()
         db_fail_msg_box.exec();
         return false;
     }
-    aDB()->disconnect(); // reset db connection to refresh layout after initial setup.
-    aDB()->connect();
```
KotzaBoss, line 124: Great that this got decoupled!

src/gui/widgets/debugwidget.cpp
```diff
@@ -169,7 +165,20 @@ void DebugWidget::on_importCsvPushButton_clicked()
...
+ * \brief Acess the database to trigger aDB()::databaseUpdated
+ */
+void DebugWidget::touchDatabase()
+{
+    QMap<QString, QVariant> debugData;
+    debugData.insert("lastname","debugLastName");
+    debugData.insert("firstname","debugFirstName");
+    auto pilot = APilotEntry(1);
+    pilot.setData(debugData);
+    aDB()->commit(pilot);
```
KotzaBoss, line 181: Isnt that going to put garbage data in the database?

src/database/adatabase.cpp
fiffty-50: It is a bit more expensive performance-wise, yes. But it adds value by being more robust. Also it only affects the commit queries, so the majority of queries are unaffected by this. Accessing data from the database is very fast, so I am not concerned about the performance loss considering that we are gaining always up-to-date data instead of keeping a (potentially out-of-date) copy of that data around in a member variable.

src/gui/widgets/debugwidget.cpp
fiffty-50: in a way, yes - it resets the logbook owner name to debug values, but since this function is called when the database is reset and filled with testing data for debugging, this is what it's supposed to do as well.

src/gui/dialogues/firstrundialog.cpp
fiffty-50: that's the benefit of having up-to date layout data around, no need to re-call conect() or update()

src/database/adatabase.cpp
fiffty-50: thinking it through again you make a good point though, the changes in layout really only happen during the initial setup so taking the performance hit throughout the application seems unreasonable. Implementing an updateLayout() function to update the member variables whenever layout changes.

KotzaBoss: That does seem like a good hybrid in our case at least we *prove* otherwise

src/gui/widgets/debugwidget.cpp
KotzaBoss: i believe it goes also to [this](https://github.com/fiffty-50/openpilotlog/pull/43#discussion_r551342980).
The "refresh database" functionality can be decoupled from all database operations. So commit calls `refresh` or whatever its called. That way debug can just call refresh which does just that.

fiffty-50: fixed

src/database/adatabase.cpp
fiffty-50: done

PR #47, opened 2021-01-05
Guard suggestions

KotzaBoss:
Guards of ASettings and AStyle is how you version check.
Use of anythings that is guarded means that the developer must be aware of it so potentially he guards the use aswell.

For not trivial problems it can look like two code blocks between the #elif with separate calls to functions.

### DIFF
--- a/src/classes/asettings.h
+++ b/src/classes/asettings.h
@@ -17,6 +17,7 @@
  */
 #ifndef ASETTINGS_H
 #define ASETTINGS_H
+#include <QtGlobal>
 #include <QtCore>
 #include <QSettings>
 
@@ -88,6 +89,10 @@ public:
     static QVariant read(const NewFlight key);
     static void write(const NewFlight key, const QVariant &val);
 
+#if QT_VERSION > QT_VERSION_CHECK(5, 9, 0)
+    static QSettings settings() { return QSettings(); }
+#endif
+
     // [G]: enum class may be making it abit too strict perhaps?
     // a workaround is to use plain enums and have one function takes an int
     // All enums should be unique of course thats easy. See At the end of the file
@@ -112,7 +117,6 @@ public:
     static QString stringOfKey(const Setup key);
     static QString stringOfKey(const UserData key);
 
-    static QSettings settings();
 
 private:
     static QMap<Main, QString> mainMap;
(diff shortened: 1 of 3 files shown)


src/classes/astyle.cpp
```diff
@@ -1,13 +1,24 @@
...
+#include <QtGlobal>
 #include <QStyleFactory>
 #include <QApplication>
 #include "src/testing/adebug.h"
 #include "src/classes/asettings.h"
 
 // this was throwing hundreds of warnings on 5.9.5, fusion is a good default for all platforms, let's
 // stick with that for now.
+// [G]: Ensure QStringLiteral is in version
+#if QT_VERSION > QT_VERSION_CHECK(5, 9, 0)  // we can set macros like STABLE_QT or smth for readability of course
```
fiffty-50, line 12: While being a good solution from a code-perspective it is not addressing the second point of my comment:

`fusion is a good default for all platforms, let's stick with that for now`
I take responsibility since I put the other reason first - But Fusion is the default out-of-the-box theme for qt on all platforms. It is universally supported by all versions of qt5 on all platforms, so it should be the default choice.

Same note on the logic as above

src/classes/asettings.h
```diff
@@ -88,6 +89,10 @@ class ASettings {
     static QVariant read(const NewFlight key);
     static void write(const NewFlight key, const QVariant &val);
 
+#if QT_VERSION > QT_VERSION_CHECK(5, 9, 0)
```
fiffty-50, line 92: What is the purpose of this function? It is not used anywhere in the application.

If we need it, we can implement it, but for now I would rather not have a function around which is:
- unused
- potentially troublesome

A note on the logic: Just because it is not good for 5.9 doesn't mean it's good for 5.10 or 5.11, we only know it's okay for 5.12 and higher.

fiffty-50 (changes requested):
I see the point you are trying to make and thank you for taking the time to type this out so I could more easily understand it.

But there is really no need to use the guards in these two cases. 

- In the first case, they are used to keep an obselete function around which is not being used anyway.
- In the second case I agree with the usefulness of the concept, it would do what it is intending to do. However, as explained in my comment, my intention is to have "Fusion" as the default theme on all platforms.

I understand the usefulness of these guards, but in these cases I am not convinced of their neccessity.

src/classes/asettings.h
KotzaBoss: I dont know why we have it so we can just remove it for now. I dont remember its use.

PR #48, opened 2021-01-05
Develop qt 5 9 5 refactor

fiffty-50:
Some minor adjustments to ensure compatibility with earlier qt versions, specifically Qt 5.9.5, which is included in Ubuntu 18.04 LTS and would be a sensible target for creating an appimage with broad compatibility (with 16.04 ending support in April 2021). 

The changes are minor and mostly related to syntax. As a general observation, I noticed that older versions of Qt seem to be less forgiving when creating QVectors, QPairs and the like when the template arguments are not explicitly specified, so we should keep the following in mind for further development:

```cpp
auto var = QPair{value, value} // not good
auto var = QPair<T,T>{value, value} // better
```

Apart from that, nothing really noteworthy, I encountered a problem with ADownload but upon investigation that was a problem in my configuration.

Misc:

- Inline removed from definition of ADatabase::instance() since it gave linker errors.
- Adjusted the .pro file to account for older nomenclature (c++1z vs c++17), and surpress unwanted warnings related to [QTBUG-68156](https://bugreports.qt.io/browse/QTBUG-68156)
- Fusion style default for all platforms

Users and developers who compile with never versions of qt should be unaffected by these changes.

### DIFF
--- a/src/gui/dialogues/newflightdialog.cpp
+++ b/src/gui/dialogues/newflightdialog.cpp
@@ -26,15 +26,6 @@
 
 #include "src/testing/adebug.h"
 
-/////////////////////////////////////// DEBUG /////////////////////////////////////////////////////
-void NewFlightDialog::onInputRejected()
-{
-    auto sender_object = sender();
-    auto line_edit = this->findChild<QLineEdit*>(sender_object->objectName());
-    DEB << line_edit->objectName() << "Input Rejected - " << line_edit->text();
-}
-/////////////////////////////////////// DEBUG /////////////////////////////////////////////////////
-
 ///////////////////////////////////////////////////////////////////////////////////////////////////
 ///                                         constants                                           ///
 ///////////////////////////////////////////////////////////////////////////////////////////////////
@@ -282,10 +273,6 @@ void NewFlightDialog::setupSignalsAndSlots()
     auto line_edits = this->findChildren<QLineEdit*>();
 
     for (const auto &line_edit : line_edits){
-        /////////////////////////////////////// DEBUG /////////////////////////////////////////////////////
-        QObject::connect(line_edit, &QLineEdit::inputRejected,
-                         this, &NewFlightDialog::onInputRejected);
-        /////////////////////////////////////// DEBUG /////////////////////////////////////////////////////
         line_edit->installEventFilter(this);
         if(line_edit->objectName().contains("Loc")){
             QObject::connect(line_edit, &QLineEdit::textChanged,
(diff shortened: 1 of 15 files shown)


src/gui/dialogues/newflightdialog.cpp
```diff
@@ -26,15 +26,6 @@
...
 
-/////////////////////////////////////// DEBUG /////////////////////////////////////////////////////
-void NewFlightDialog::onInputRejected()
-{
-    auto sender_object = sender();
-    auto line_edit = this->findChild<QLineEdit*>(sender_object->objectName());
-    DEB << line_edit->objectName() << "Input Rejected - " << line_edit->text();
-}
-/////////////////////////////////////// DEBUG /////////////////////////////////////////////////////
-
```
KotzaBoss, line 37: Maybe trim the //// a bit

src/gui/widgets/debugwidget.cpp
```diff
@@ -148,24 +149,32 @@ void DebugWidget::on_importCsvPushButton_clicked()
...
-
+#if QT_VERSION >= QT_VERSION_CHECK(5, 12, 0)
+    DEB << "SSL version use for build: " << QSslSocket::sslLibraryBuildVersionString();
+    DEB << "SSL version use for run-time: " << QSslSocket::sslLibraryVersionNumber();
+    DEB << QT_VERSION_MAJOR << QT_VERSION_MINOR << "At least 5.12";
+#else
+    DEB << "SSL version use for build: " << QSslSocket::sslLibraryBuildVersionString();
+    DEB << "SSL version use for run-time: " << QSslSocket::sslLibraryVersionNumber();
+    DEB << QT_VERSION_MAJOR << QT_VERSION_MINOR << "Less than 5.12";
+#endif
```
KotzaBoss, line 177: since the first two lines are identical this could be written as
```c++
DEB << "SSL version use for build: " << QSslSocket::sslLibraryBuildVersionString();
DEB << "SSL version use for run-time: " << QSslSocket::sslLibraryVersionNumber();
#if  QT_VERSION >= QT_VERSION_CHECK(5, 12, 0)
DEB << QT_VERSION_MAJOR << QT_VERSION_MINOR << "At least 5.12";
#else
DEB << QT_VERSION_MAJOR << QT_VERSION_MINOR << "Less than 5.12";
#endif
```

src/gui/dialogues/newflightdialog.cpp
fiffty-50: that block is removed, but I'll keep it in mind for the future!

src/gui/widgets/debugwidget.cpp
fiffty-50: actually I forgot to clean that, it was for debugging an issue I had. Removing.
Thanks for the feedback!

PR #54, opened 2021-01-06
Fix for bug #51

### LINKED ISSUE #51: : is an illegal character in file names for windows
Upon testing a windows build I came across this problem preventing FirstRunDialog from finishing correctly. Leaving this here because I don't have access to git on this machine.

```cpp
bool ADataBaseSetup::backupOldData()
{
    auto database_file = aDB->databaseFile;
    if(!database_file.exists()) {
        DEB << "No Database to backup, returning.";
        return true;
    }

    auto date_string = QDateTime::currentDateTime().toString("yyyy_MM_dd_T_hh_mm");
    auto backup_dir = QDir(AStandardPaths::absPathOf(AStandardPaths::DatabaseBackup));
    auto backup_name = database_file.baseName() + "_bak_" + date_string + ".db";
    auto file = QFile(aDB->databaseFile.absoluteFilePath());

    if (!file.rename(backup_dir.absolutePath() + '/' + backup_name)) {
        DEB << "Unable to backup old database.";
        return false;
    }
    DEB << "Backed up old database as: " << backup_name;
    return true;
}

```

fiffty-50:
Fixed a bug that was putting a not cross-platform safe character into a filename.

Added a cast to avoid container detachment ref https://doc.qt.io/qt-5/qtglobal.html#qAsConst

closes #51

### DIFF
--- a/src/functions/acalc.h
+++ b/src/functions/acalc.h
@@ -2,6 +2,7 @@
 #define ACALC_H
 
 #include <QDateTime>
+#define _USE_MATH_DEFINES // needed for MSVC
 #include <cmath>
 #include <QDebug>
 /*!
--- a/src/classes/astyle.h
+++ b/src/classes/astyle.h
@@ -13,9 +13,9 @@ class AStyle
 {
 public:
     enum StyleSheet{
-        Dark = 0,
+        Default = 0,  // Nothing read from config file produces .UInt() == 0
+        Dark,
         Light,
-        Default,
     };
 private:
     static QString currentStyle;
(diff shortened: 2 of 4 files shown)


src/functions/acalc.h
```diff
@@ -2,6 +2,7 @@
 #define ACALC_H
 
 #include <QDateTime>
+#define _USE_MATH_DEFINES // needed for MSVC
```
KotzaBoss, line 5: If only one compiler needs this it should be in a guard
https://stackoverflow.com/questions/5850358/is-there-a-preprocessor-define-that-is-defined-if-the-compiler-is-msvc
```c++
#if _MSC_VER
...
#endif

Comment (fiffty-50) [reply src/functions/acalc.h]: fixed

Comment (KotzaBoss) [reply src/functions/acalc.h]: *CORRECTION*
```c++
#ifdef _MSC_VER
...
#endif
```

Issue #53, opened 2021-01-06
Default Style Sheet Dark


fiffty-50:
Observed behaviour:

On first run, the FirstRunDialog is styled with the dark stylesheet

Desired behaviour:

Default stylesheet should be no style shee.

Observed on: Windows 10, Qt 5.15.2

Steps to reproduce: Delete all config files and start the application.

Additional Observation: Fusion is a good choice for windows, the 'native' windows style looks horrible.

KotzaBoss:
Confirm commit 82dda30e515223a794b5145ace3b7c21f509a277 works as requested

Issue #52, opened 2021-01-06
Work required on ADownload class/FirstRunDialog


fiffty-50:
The ADownload class depends on a working openSSL installation in order to properly download the templates.

With the current behaviour, if a connection cannot be established, the download finished signal is triggered and the result written as an empty file. This is a problem in general, but more so in the FirstRunDialog, where overwriting the template csv files renders template database creation inop and leaves the user stuck with a failing FirstRunDialog.

The ADownload class needs guards not to overwrite local files unless a connection was successfully established.
We should also harden the FirstRunDialog, I am thinking about seperating the 'default' templates from the 'downloaded' templates, in order to always have a working fallback in case errors occur during the download.

It might even be worth considering to always use local data for the database templates in the firstrundialog and add an option to "update" the database at a later stage from the gui. Like this, even if an error occurs, the user still has a working application.

PR #55, opened 2021-01-06
Merging issue closing #52

fiffty-50:


src/database/adatabasesetup.cpp
```diff
@@ -302,12 +302,12 @@ bool ADataBaseSetup::backupOldData()
         return true;
     }
 
-    auto date_string = QDateTime::currentDateTime().toString(Qt::ISODate);
+    auto date_string = QDateTime::currentDateTime().toString("yyyy_MM_dd_T_hh_mm");
```
KotzaBoss, line 305: From 20b697d14ea257b985a6292a72b519651ac5d7d1
```c++
auto date_string = QDateTime::currentDateTime().toString(Qt::ISODate);
#if defined(_WIN32) || defined(_WIN64)
date_string.replace(':', '-');
#endif
```

fiffty-50: better to keep the filename consistent across platforms

src/database/adatabasesetup.cpp
```diff
@@ -302,12 +302,12 @@ bool ADataBaseSetup::backupOldData()
...
     auto backup_dir = QDir(AStandardPaths::absPathOf(AStandardPaths::DatabaseBackup));
-    auto backup_name = database_file.baseName() + "_bak_" + date_string + ".db";
+    auto backup_name = database_file.baseName() % "_bak_" % date_string + ".db";
```
KotzaBoss, line 307: https://doc.qt.io/qt-5/qstring.html#more-efficient-string-construction
We can keep the fairly known `+` operator for concatenation but use qmake to our advantage for efficient string building

fiffty-50: i tried that, unfortunately using qstringbuilder is not compatible with the whole project at the moment

src/database/adatabasesetup.cpp
KotzaBoss: Then we need our custon function to be used instead of `.toString(...)` because hardcoding it that way isnt preferable
```c++
QString dateToString(QDateTime& dt) {
    ...
    return custom_time_string
}
```

fiffty-50: Why add overhead by writing a custom function for `.toString()` that does the same as `.toString()` but in a non-standard and probably less efficient way?

What would that function do other than

```cpp
QString dateToString(QDateTime& dt) {
    auto custom_time_string = dt.toString("yyyy_MM_dd_T_hh_mm");
    return custom_time_string
}
```
?

If it does only that, why add the overhead? If it does something more, why would that be required?

KotzaBoss: because we control the output to and from (if needed) in a form way. The "overhead" is negligible, the readability inscreased and we can give the compiler hint to inline.
We pack the custom logic of reading and writing the backup file string into 2 functions that hide any .toString(), str.replace()

fiffty-50: For dates this is no issue, since isodate does not include the time, only datetime does.
`date.toString(Qt::ISODate)` returns `yyyy-MM-dd`

For reading the backup files, checking the filename is a flawed approach, since:

1. We cannot control how the user may rename his backups.
2. The file name says nothing about the validity of its contents.

I agree that the overhead would be negligible, but I prefer the simplicity of using the function that Qt provides over duplicating its functionality and employing custom logic.

QDate and QDateTime, date and time operations are immensely complex. Since Qt has a function that does exactly what we need (output the current date and time in a format we control) - `.toString()` I prefer not to include custom logic that does little more than duplicating what is already included in Qt.

However, I do agree with you that hardcoding a string here is not the perfect solution. I have added a namespace that collects constants that are used throughout the application - this will ensure uniformity of date, time and datetime formats and adresses the isssue with this function.

See [19232fa](https://github.com/fiffty-50/openpilotlog/pull/55/commits/19232fa5981e73e0d42b39a3aabcc075ffe0c328)

src/oplconstants.h
```diff
@@ -0,0 +1,113 @@
...
+
+/*!
+ * The datetime namespace specifies date and time formats by QDate, QTime
+ * and QDateTime that deviate from standard values included in the Qt Framework
+ * like Qt::ISODate and are to be used in conjunction with the .toString() members
+ * of these classes
+ */
+namespace datetime {
+
+static const auto TIME_DEFAULT_FORMAT    = QStringLiteral("hh:mm");
```
KotzaBoss, line 19: Can just use Qt::TextDate for uniformity.
https://doc.qt.io/qt-5/qtime.html#toString-1

fiffty-50: Qt::ISODate returns HH:mm:ss, Flight Times are always logged only to minutes in accuracy. Since we are not logging, using or displaying seconds, Qt::ISODate is unsuitable.

KotzaBoss: Then since this is a domain specific thing it should be named as such.
FLIGHT_TIME_FORMAT seems more suitable

fiffty-50: Hm then we would another one for the general default. It also applies to departure time, arrival time etc. But having FLIGHT_TIME_DEFAULT_FORMAT and TIME_DEFAULT_FORMAT is not a bad idea, since we may include[additional flight time formats](https://github.com/fiffty-50/openpilotlog/wiki/Future-Features-ideas#3-support-for-decimal-time-support) in the future. I'll change it.

src/functions/acalc.h
```diff
@@ -44,6 +49,9 @@ inline QTime blocktime(QTime tofb, QTime tonb)
  */
 inline QString minutesToString(QString block_minutes)
 {
+    DEB << "#######################################";
+    DEB << "# WARNING: USE OF DEPRECATED FUNCTION #";
+    DEB << "#######################################";
```
KotzaBoss, line 54: We can avoid polluting debug out put with a simple deprecation macro
```c++
QT_DEPRECATED
inline QSTring minutesToString(...)
```

src/functions/adatetime.h
```diff
@@ -0,0 +1,24 @@
...
+ * \brief toString formats a QDateTime object into a string in a uniform way.
+ * \return
+ */
+inline const QString toString (const QDateTime date_time, opl::datetime::DateTimeFormat format) {
+    switch (format) {
+    case opl::datetime::Default:
+        return date_time.toString(Qt::ISODate);
+    case opl::datetime::Backup:
+        return date_time.toString(QStringLiteral("yyyy_MM_dd_T_hh_mm"));
+        break;
```
KotzaBoss, line 18: no need to break if you return. Compiler may whine as well

src/functions/acalc.h
fiffty-50: I added the QT_DEPRECATED makro, very helpful!

src/functions/adatetime.h
fiffty-50: fixed

src/functions/atime.h
```diff
@@ -48,7 +47,7 @@ inline const QString minutesToString(int minutes_in, opl::time::FlightTimeFormat
     {
         int hour = minutes_in / 60;
         double minute = (minutes_in % 60) / 60.0;
-        return QString::number(hour+minute);
+        return QString::asprintf("%.2f", (hour+minute));
```
KotzaBoss, line 50: Qt recommends arg for future code. https://doc.qt.io/qt-5/qstring.html#asprintf.
```c++
return QString::number("%1").arg(hour + minute, 0, 'f', 2);
```

fiffty-50: Good idea, on it!

fiffty-50:
closes #51 and closes #53

PR #57, opened 2021-01-12
Guards against unsuccessful Downloads in FirstRunDialog

### LINKED ISSUE #52: Work required on ADownload class/FirstRunDialog

fiffty-50:
The templates that are used to create the initial database are now included in the project itself. This enables the user to choose not to download the latest revision should there be connectivity issues.

Even if the user chooses to download the latest templates, there is now a fallback mechanism to create the database from local data, in case the download fails. This leaves the user with a working application instead of being stuck at the FirstRunDialog.

Creating the database layout has been decoupled from filling in the template data.

This commit closes #52

### DIFF
--- a/src/classes/aaircraftentry.h
+++ b/src/classes/aaircraftentry.h
@@ -1,3 +1,20 @@
+/*
+ *openPilot Log - A FOSS Pilot Logbook Application
+ *Copyright (C) 2020  Felix Turowsky
+ *
+ *This program is free software: you can redistribute it and/or modify
+ *it under the terms of the GNU General Public License as published by
+ *the Free Software Foundation, either version 3 of the License, or
+ *(at your option) any later version.
+ *
+ *This program is distributed in the hope that it will be useful,
+ *but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *GNU General Public License for more details.
+ *
+ *You should have received a copy of the GNU General Public License
+ *along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
 #ifndef AAIRCRAFTENTRY_H
 #define AAIRCRAFTENTRY_H
 
--- a/src/classes/astandardpaths.cpp
+++ b/src/classes/astandardpaths.cpp
@@ -1,30 +1,33 @@
 #include "src/classes/astandardpaths.h"
 
-QMap<AStandardPaths::Dirs, QString> AStandardPaths::dirs;
+QMap<AStandardPaths::Directories, QString> AStandardPaths::directories;
 
 void AStandardPaths::setup()
 {
     auto data_location = QStandardPaths::AppDataLocation;
-    dirs = {  // [G]: potential rename to `dirs`
-        {Database, QStandardPaths::writableLocation(data_location)},
-        {Templates, QDir(QStandardPaths::writableLocation(data_location)).filePath("templates")},
-        {DatabaseBackup, QDir(QStandardPaths::writableLocation(data_location)).filePath("backup")}
+    directories = { // [F]: Dir could be ambiguous since it is very similar to QDir
+        {Database, QDir::toNativeSeparators(
+         QStandardPaths::writableLocation(data_location) + '/')},
+        {Templates, QDir::toNativeSeparators(
+         QDir(QStandardPaths::writableLocation(data_location)).filePath(QStringLiteral("templates/")))},
+        {DatabaseBackup, QDir::toNativeSeparators(
+         QDir(QStandardPaths::writableLocation(data_location)).filePath(QStringLiteral("backup/")))}
     };
 }
 
-const QString& AStandardPaths::absPathOf(Dirs loc)
+const QString& AStandardPaths::absPathOf(Directories loc)
 {
-    return dirs[loc];
+    return directories[loc];
 }
 
-const QMap<AStandardPaths::Dirs, QString>& AStandardPaths::allPaths()
+const QMap<AStandardPaths::Directories, QString>& AStandardPaths::allPaths()
 {
-    return dirs;
+    return directories;
 }
 
 void AStandardPaths::scan_dirs()
 {
-    for(auto& dir : dirs){
+    for(auto& dir : directories){
         auto d = QDir(dir);
         if(!d.exists()) {
             d.mkpath(dir);
@@ -32,9 +35,14 @@ void AStandardPaths::scan_dirs()
     }
 }
 
+// [F]: We could make scan_dirs return bool and return false if !exists && !mkpath
+// This function seems like it would do the same as scan_dirs
+// Validating the contents would be rather complex to accomplish and difficult to
+// maintain I believe, since the contents of these directories might change and it
+// seems out of the scope of this class to verify the directories' contents.
 bool AStandardPaths::validate_dirs()
 {
-    for(auto& dir : dirs){
+    for(auto& dir : directories){
         //DEB << "Validating " << dir;
         if(false)  // determine path as valid (scan contents and parse for correctness)
             return false;
--- a/src/classes/astandardpaths.cpp
+++ b/src/classes/astandardpaths.cpp
@@ -1,51 +1,37 @@
 #include "src/classes/astandardpaths.h"
 
-QMap<AStandardPaths::Directories, QString> AStandardPaths::directories;
+QMap<AStandardPaths::Directories, QDir> AStandardPaths::directories;
 
 void AStandardPaths::setup()
 {
     auto data_location = QStandardPaths::AppDataLocation;
     directories = { // [F]: Dir could be ambiguous since it is very similar to QDir
-        {Database, QDir::toNativeSeparators(
-         QStandardPaths::writableLocation(data_location) + '/')},
-        {Templates, QDir::toNativeSeparators(
-         QDir(QStandardPaths::writableLocation(data_location)).filePath(QStringLiteral("templates/")))},
-        {DatabaseBackup, QDir::toNativeSeparators(
-         QDir(QStandardPaths::writableLocation(data_location)).filePath(QStringLiteral("backup/")))}
+        {Database, QDir(QStandardPaths::writableLocation(data_location) + '/')},
+        {Templates, QDir(QStandardPaths::writableLocation(data_location)).filePath(
+         QStringLiteral("templates/"))},
+        {Backup, QDir(QStandardPaths::writableLocation(data_location)).filePath(
+         QStringLiteral("backup/"))}
     };
 }
 
-const QString& AStandardPaths::absPathOf(Directories loc)
+const QDir& AStandardPaths::directory(Directories loc)
 {
     return directories[loc];
 }
 
-const QMap<AStandardPaths::Directories, QString>& AStandardPaths::allPaths()
+const QMap<AStandardPaths::Directories, QDir>& AStandardPaths::allDirectories()
 {
     return directories;
 }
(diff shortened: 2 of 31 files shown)


src/database/adatabasesetup.cpp
```diff
@@ -328,13 +327,26 @@ bool ADataBaseSetup::importDefaultData()
...
+            }
+        } else {
+            DEB << "Using downloaded resources";
+            if (!commitData(aReadCsv(AStandardPaths::absPathOf(AStandardPaths::Templates)
+                                     % QLatin1Char('/')
+                                     % table % QStringLiteral(".csv")),
+                            table)) {
+                DEB << "Error importing data (remote).";
+                return false;
+            }
```
KotzaBoss, line 349: This screams code duplications.
```c++
// if `/` is necessary then AStandardPaths should have it by default
// we should see how ".../directory/" is handled in general
// i assume Qt just ignores the last `/`

// Outside for loop
QString data_to_commit_path;
QString error_msg = "Error importing data ";
if(use_local_data){
	data_to_commit_path = ":templates/database/templates/";
	error_msg = "(local)";
} 
else {
	data_to_commit_path = AStandardPaths::absPathOf(AStandardPaths::Templates)
	error_msg = "(remote)";
}

// In for loop
if(!commitData(data_to_commit + table + ".csv", table)){
	DEB << error_msg;
	return false;
}
return true;
```
if it is expecting a file with name `table.csv` then why does `commitData` take it as a second parameter?

src/database/adatabasesetup.cpp
```diff
@@ -292,6 +286,11 @@ bool ADataBaseSetup::downloadTemplates()
         dl->setFileName(template_dir.filePath(table % QStringLiteral(".csv")));
         dl->download();
         loop.exec(); // event loop waits for download done signal before allowing loop to continue
+
+        QFileInfo downloaded_file(template_dir.filePath(table % QStringLiteral(".csv")));
+        if (downloaded_file.size() == 0)
+            return false; // ssl/network error
+
```
KotzaBoss, line 293: Are we sure there is no leak here?

src/database/adatabasesetup.cpp
fiffty-50: #### Standard Paths
The strings provided by QStandardPaths do not include a trailing `/` character, as can be seen [here](https://doc.qt.io/qt-5/qstandardpaths.html#StandardLocation-enum). We need to add it when writing to a file, because otherwise what would happen is a file like this would be created:
`path/to/dirfile.ini`
instead of
`path/to/dir/file.ini`

I am not sure about the reason behind this, for me it would make sense to append a trailing `/` character to indicate we are referencing a directory, since without it, it may be ambiguous in some contexts. Adding the trailing separator to AStandardPaths might bring with it some complication with regard to cross-platform compatibility, QDir`s [toNativeSeperators](https://doc.qt.io/qt-5/qdir.html#toNativeSeparators) might be useful. Should we go for a solution like that we should document it clearly, since it is 'non-standard' behaviour for Qt - I'll look into it.
#### commitData / loop
`commitData()` takes the parsed data to be committed as a first argument, and the table it is supposed to commit it to as the second argument. These look like the same in this case, because the file is named like the table, so `aReadCsv()` and `commitData()` take the same argument, but that is not always the case.
I will rewrite it a little to make that distinction more visible, including you suggestions on the loop!

KotzaBoss: Then we should rework AStandardPaths to be more robust. QStrings may be too generic and clunky for the specific use paths are. Since standard paths are directories (if we want standard files we can extend it of course) we could have them like this.
```c++
    dirs = {  // [G]: potential rename to `dirs`
        {Database, QDir(QStandardPaths::writableLocation(data_location))},
        {Templates, QDir(QStandardPaths::writableLocation(data_location) + "/templates")},
        {DatabaseBackup, QDir(QStandardPaths::writableLocation(data_location) + "/backup")}
    };
```
This seems like the simple way to construct the new path and i dont see any reason to overcomplicate it.
Making AStandardPaths QDirs instead of QString means that we can use the existing methods that deal with the plumbing of slashes and so on.
The above can be used as follows:
```c++
...
// more robust and safe since Qt deals with the path plumbing
if(!commitData(data_to_commit_dir.filePath(table + ".csv"), table)){ 
	DEB << error_msg;                                                                   
	return false;
}
```
In order to refactor `commitData` we can start by overloaing it first so that it deals with `QDir` logic and incrementaly delete the `QString` oveload if it is not needed.

src/classes/astandardpaths.cpp
```diff
@@ -1,40 +1,48 @@
...
-        {Database, QStandardPaths::writableLocation(data_location)},
-        {Templates, QDir(QStandardPaths::writableLocation(data_location)).filePath("templates")},
-        {DatabaseBackup, QDir(QStandardPaths::writableLocation(data_location)).filePath("backup")}
+    directories = { // [F]: Dir could be ambiguous since it is very similar to QDir
+        {Database, QDir::toNativeSeparators(
+         QStandardPaths::writableLocation(data_location) + '/')},
+        {Templates, QDir::toNativeSeparators(
+         QDir(QStandardPaths::writableLocation(data_location)).filePath(QStringLiteral("templates/")))},
+        {DatabaseBackup, QDir::toNativeSeparators(
+         QDir(QStandardPaths::writableLocation(data_location)).filePath(QStringLiteral("backup/")))}
```
KotzaBoss, line 14: That... doesnt look sane. Qt wants `/` for all separators nomatter what system. The toNative just returns the OS dependant separators and it should be used at the edges of the logic that needs paths. AStandardPaths should have the general QDirs that other parts of the program tweak to their specific use. Having already nativeSeparators in the `directories` means new paths need to have the nativeSeparator called as well and we are starting again with clunky code.

fiffty-50: That's reasonable, preparing a commit!

src/database/adatabasesetup.cpp
fiffty-50: Yes, since they are created before the main event loop is started and marked for deletion with deleteLater(), they are deleted at this point, see [here](https://doc.qt.io/qt-5/qobject.html#deleteLater).

src/classes/astandardpaths.cpp
```diff
@@ -1,51 +1,37 @@
...
-         QStandardPaths::writableLocation(data_location) + '/')},
-        {Templates, QDir::toNativeSeparators(
-         QDir(QStandardPaths::writableLocation(data_location)).filePath(QStringLiteral("templates/")))},
-        {DatabaseBackup, QDir::toNativeSeparators(
-         QDir(QStandardPaths::writableLocation(data_location)).filePath(QStringLiteral("backup/")))}
+        {Database, QDir(QStandardPaths::writableLocation(data_location) + '/')},
+        {Templates, QDir(QStandardPaths::writableLocation(data_location)).filePath(
+         QStringLiteral("templates/"))},
+        {Backup, QDir(QStandardPaths::writableLocation(data_location)).filePath(
+         QStringLiteral("backup/"))}
```
KotzaBoss, line 13: Quick: `/` are redundant.

This code is suspect. `QDir("...").filePath("...")` returns a QString however a QDir with *that* string is created. The QString is somehow implicitely used for the QDir construction. I still believe my original idea is still the simplest explicit one
```c++
{Database, QDir(QStandardPaths::writableLocation(data_location))},  // no / needed
{Templates, QDir(QStandardPaths::writableLocation(data_location) + QStringLiteral("/templates"))},  // here it is at the start
{Backup, QDir(QStandardPaths::writableLocation(data_location) + QStringLiteral("/backup"))}
```
The QDir is explicitelly created through its consructor with no funny business

src/classes/astandardpaths.cpp
```diff
@@ -1,51 +1,37 @@
...
-void AStandardPaths::scan_dirs()
+bool AStandardPaths::scan_dirs()
 {
-    for(auto& dir : directories){
-        auto d = QDir(dir);
-        if(!d.exists()) {
-            d.mkpath(dir);
+    for(const auto& dir : directories){
+        if(!dir.exists()) {
+            DEB << dir << "Does not exist. Creating: " << dir.absolutePath();
```
KotzaBoss, line 31: Secondary note: qDebug() leaves a space between the `<<` calls.
```c++
DEB << "abc" << "d";  // abc d
```

src/classes/astandardpaths.cpp
fiffty-50: I'm gonna be honest with you, that went over my head.

But I see the benefits of your approach, commit coming up!

src/classes/aaircraftentry.h
```diff
@@ -1,3 +1,20 @@
...
+ *(at your option) any later version.
+ *
+ *This program is distributed in the hope that it will be useful,
+ *but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *GNU General Public License for more details.
+ *
+ *You should have received a copy of the GNU General Public License
+ *along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
```
KotzaBoss, line 17: Putting the license on every file is just too much.

fiffty-50: It's a [requirement](https://www.gnu.org/licenses/gpl-faq.html#NoticeInSourceFile) for the GPL license to be included in every source file. It is not the whole license but the [preamble](https://www.gnu.org/licenses/gpl-faq.html#GPLOmitPreamble) which must not be omitted.

KotzaBoss: Nothing in both links says that you have to put the license/preaamble in every file.

The first link says that every file should have a *notice* of the license its under. Havent seen it specified, could be just one line stating that the current file is under GPL license for the case someone separates it from the source.

The second link says that the *license* should not omit *any* parts of the gpl.

fiffty-50: > Havent seen it specified,

If you read on a bit more on the page, you will find explanations going more into details on how those terms are to be interpreted.

You can find an explanation on the format of the notice [here](https://www.gnu.org/licenses/gpl-howto.html) under the header license notices.

That this license notice (and the copyright notice) can not be omitted is stated in the first link.

KotzaBoss: This is ludicrous... why pollute every file with 40 lines. I get the "separated from source" part but cmon

fiffty-50: It's not ideal, but I don't make these rules, we'll just have to comply with it.

I chose GPL because Qt is a commercial framework. It can be used under open source licensing, but there are certain [restrictions](https://www.qt.io/licensing/) and the project using Qt under an open source license needs to be either GPL or LGPL licensed. As for what has to be included in the license notice, there is no difference between the two, but the GPL is more expansive and detailed, which leads to using Qt under the GPL having [less restrictions](https://www.qt.io/licensing/open-source-lgpl-obligations#lgpl).

PR #58, opened 2021-01-18
Develop string rework

### LINKED ISSUE #56: openPilotLog not compatible with QStringBuilder globally
Qt offers [QStringBuilder](https://doc.qt.io/qt-5/qstring.html#more-efficient-string-construction) to efficiently create strings from concatenations using the `%` operator to replace `+`.

If `DEFINES *= QT_USE_QSTRINGBUILDER` is in the .pro file, the `+` operator can be used in the source code and will automatically interpreted as `%` where appropriate, making its use very convenient. Unfortunately this solution is not entirely source compatible, and currently not possible for this project.

Using the `%` operator is counter-intuitive and not a widely known way of concatenating strings, so it would be desirable to be able to enable QStringBuilder globally and keep using the `+` operator.

In order to enable QStringBuilder globally, we would need to refactor the incompatible code. At a first view these seem to be the parts creating problems:

- REPR makro
- ADatabaseError

fiffty-50:
closes #56

- Added tr() for all user-facing strings
>In order for translations to be possible at a later stage, tr() should be used for all user-facing strings. 
- Enabled QStringbuilder globally and edited wiki with [guidance](https://github.com/fiffty-50/openpilotlog/wiki/Coding-Style#strings-in-openpilotlog)


src/gui/dialogues/newflightdialog.cpp
```diff
@@ -49,10 +49,10 @@ static const auto SELF_RX              = QRegularExpression(
...
 
 static const auto MANDATORY_LINE_EDITS_DISPLAY_NAMES = QMap<int, QString> {
-    {0, QStringLiteral("Date of Flight")}, {1, QStringLiteral("Departure Airport")},
-    {2, QStringLiteral("Destination Airport")}, {3, QStringLiteral("Time Off Blocks")},
-    {4, QStringLiteral("Time on Blocks")}, {5, QStringLiteral("PIC Name")},
-    {6, QStringLiteral("Aircraft Registration")}
+    {0, QObject::tr("Date of Flight")},      {1, QObject::tr("Departure Airport")},
+    {2, QObject::tr("Destination Airport")}, {3, QObject::tr("Time Off Blocks")},
+    {4, QObject::tr("Time on Blocks")},      {5, QObject::tr("PIC Name")},
+    {6, QObject::tr("Aircraft Registration")}
```
KotzaBoss, line 55: This looks like it needs an enum to keep track of manual accesses.
Even if we keep the current organisations with the separate QBitArray for verifying validity its best to be proactive about it.
```c++
enum LineEditDispName {
    DateOfFlight, DestAirport, ...
}
```
Newflightsdialogs bookeeping can be cleaned up abit aswell. To minimise the refactoring we can just move the mandatoryLineEdits and the mandatoryLineEditsGood to a struct. I will play around with the names bellow for simplicity's sake. I will also ignore private/public because i have to study the relations further but thats not the bulk of the refactoring
```c++
class NewFlightDIalog {
    enum LineEdits {
        DateOfFlight = 0, DestAirport, 
    };

    struct MandatoryLineEdits {
      QVector<QLineEdit*> lineEdits;
      QBitArray ok;

      void unvalidate(QLineEdit* le);
      void unvalidate(LineEdits le);

      bool operator[] (LineEdits le);
    } mandatoryLineEdits;
};

void setupRawInputValidation() {
   ...
   mandatoryLineEdits.lineEdits = {  // the array is populated with the line edit ptrs
        ui->...,
   };
   mandatoryLineEdits.ok.fill(false);  // the ok bit array is filled with false
}
```
This needs some work to be set up (which i can do if you are not intersted because i am) but the idea is that we "hide" the integer which is used to index the arrays behind an enum giving us semantic strength.
```c++
mandatoryLineEditsGood.setBit(mandatoryLineEdits.indexOf(line_edit), false);  // I dont care about that...
mandatoryLineEdits.unvalidate(line_edit);  // thats what care about at this level
```

fiffty-50: You're making good points!

But before I comment on your suggestions, let's get into organisation and workflow for a moment.
If you want to work on the NewFlightDialog that's great! But let's be mindful not to mix things up too much. This branch/commit is about adding tr() for translations and enabling QStringBuilder, so refactoring the Dialog is outside the scope of this pull request.

Here's how I would like the workflow to look like: If you identify a problem or something you would like to work on, either start a discussion or open an issue. If it is about something specific, I would prefer an issue since it makes it easier to track, and issues can be linked to and closed with commits and/or merge requests. For general discussions that won't result in code changes at the short hand, a discussion is more appropriate. If you want, just copy/paste the above comment into a new issue and we take it from there!

- Assign yourself to the issue, let me know what you'll be working on and start a branch named after the issue/discussion.

Then we can either discuss on the issue page or on the commits/pull requests. Like this we have everything organized and the info for one issue can later easily be found.

As for your suggestions, I think they are good ideas!

<details>
 <summary>Spoiler warning</summary>
  
I think moving the LE validation into a struct is sound.

I would be a bit more careful about changing the map you quoted into an enum - it is being used to present strings to the user in a messagebox and the integer corresponds to a bit in the QBitArray, so changing that would complicate things, I would suggest focussing on the struct itself first.
  
</details>

Are you happy with the translations/QStringBuilder side of things? Once this PR is merged, I'd be happy for you to work on the NewFlightDialog!

KotzaBoss: Your tr seems good. Once you are done with this ill work on a pull request for what i mentioned above. Ill tackle the line edits first and then move to the map.

For the map i wasnt talking about changing *it* to a map but enccapsulating the data it holds better so it is more clear.

PR #61, opened 2021-01-23
Initial draft or giving semantics to raw QStrings and ints.

### LINKED ISSUE #59: Simplification of NewFlightDialog
NewFlightDialog could use some work.

- Move the mandatoryLineEdits and the mandatoryLineEditsGood to a struct
- Clear up naming of variables

KotzaBoss:
closes #59 
 
In an attempt to give semantic meaning to bare QStrings and ints i propose aliasing these types with names suited for the program.

## Proposal
`using [semantic_name]_[aliased_type] = [aliased_type];`
While the names can be debated to achieve developer clarity, the idea is that those aliases hold the semantic meaning of the type and  after `_` the mechanical meaning.

### Examples
```c++
using TableName_qstr = QString;
using RowId_i = int;
using RgstrNum_i8 = int8_t;

using PilotKey_qstr = QString;
using PIlotKeys_qvec = QVector<QString>
```

### DIFF
--- a/src/database/adatabase.cpp
+++ b/src/database/adatabase.cpp
@@ -497,7 +497,7 @@ const QStringList ADatabase::getCompletionList(ADatabaseTarget target)
 }
 
 const
-QMap<ADatabase::QueryResult_qstr, ADatabase::QueryResult_i> ADatabase::getIdMap(ADatabaseTarget target)
+QMap<ADatabase::QString, ADatabase::RowId_t> ADatabase::getIdMap(ADatabaseTarget target)
 {
     QString statement;
 
@@ -538,7 +538,7 @@ QMap<ADatabase::QueryResult_qstr, ADatabase::QueryResult_i> ADatabase::getIdMap(
     }
 
     // QVector<QString> query_result;  // [G]: unused
-    auto id_map = QMap<QueryResult_qstr, QueryResult_i>();
+    auto id_map = QMap<QString, RowId_t>();
     while (query.next()) {
         id_map.insert(query.value(1).toString(), query.value(0).toInt());
     }
--- a/src/database/adatabase.cpp
+++ b/src/database/adatabase.cpp
@@ -35,7 +35,7 @@ ADatabase* ADatabase::self = nullptr;
 /*!
  * \brief Return the names of a given table in the database.
  */
-ColumnNames ADatabase::getTableColumns(TableName table_name) const
+ColumnNames_t ADatabase::getTableColumns(TableName_t table_name) const
 {
     return tableColumns.value(table_name);
 }
@@ -43,7 +43,7 @@ ColumnNames ADatabase::getTableColumns(TableName table_name) const
 /*!
  * \brief Return the names of all tables in the database
  */
-TableNames ADatabase::getTableNames() const
+TableNames_t ADatabase::getTableNames() const
 {
     return tableNames;
 }
@@ -59,7 +59,7 @@ void ADatabase::updateLayout()
 
     tableColumns.clear();
     for (const auto &table_name : tableNames) {
-        ColumnNames table_columns;
+        ColumnNames_t table_columns;
         QSqlRecord fields = db.record(table_name);
         for (int i = 0; i < fields.count(); i++) {
             table_columns.append(fields.field(i).name());
@@ -356,12 +356,12 @@ bool ADatabase::insert(AEntry new_entry)
 
 }
 
-RowData ADatabase::getEntryData(DataPosition data_position)
+RowData_t ADatabase::getEntryData(DataPosition data_position)
 {
     // check table exists
     if (!getTableNames().contains(data_position.tableName)) {
         DEB << data_position.tableName << " not a table in the database. Unable to retreive Entry data.";
-        return RowData();
+        return RowData_t();
     }
 
     //Check Database for rowId
@@ -377,14 +377,14 @@ RowData ADatabase::getEntryData(DataPosition data_position)
         DEB << "SQL error: " << check_query.lastError().text();
         DEB << "Statement: " << statement;
         lastError = check_query.lastError().text();
-        return RowData();
+        return RowData_t();
     }
 
     check_query.next();
     if (check_query.value(0).toInt() == 0) {
         DEB << "No Entry found for row id: " << data_position.rowId;
         lastError = ADatabaseError("Database entry not found.");
-        return RowData();
+        return RowData_t();
     }
 
     // Retreive TableData
@@ -401,11 +401,11 @@ RowData ADatabase::getEntryData(DataPosition data_position)
         DEB << "SQL error: " << select_query.lastError().text();
         DEB << "Statement: " << statement;
         lastError = select_query.lastError().text();
-        return RowData();
+        return RowData_t();
     }
 
     select_query.next();
-    RowData entry_data;
+    RowData_t entry_data;
 
     for (const auto &column : getTableColumns(data_position.tableName)) {
         entry_data.insert(column, select_query.value(column));
@@ -420,28 +420,28 @@ AEntry ADatabase::getEntry(DataPosition data_position)
     return entry;
 }
 
-APilotEntry ADatabase::getPilotEntry(RowId row_id)
+APilotEntry ADatabase::getPilotEntry(RowId_t row_id)
 {
     APilotEntry pilot_entry(row_id);
     pilot_entry.setData(getEntryData(pilot_entry.getPosition()));
     return pilot_entry;
 }
 
-ATailEntry ADatabase::getTailEntry(RowId row_id)
+ATailEntry ADatabase::getTailEntry(RowId_t row_id)
 {
     ATailEntry tail_entry(row_id);
     tail_entry.setData(getEntryData(tail_entry.getPosition()));
     return tail_entry;
 }
 
-AAircraftEntry ADatabase::getAircraftEntry(RowId row_id)
+AAircraftEntry ADatabase::getAircraftEntry(RowId_t row_id)
 {
     AAircraftEntry aircraft_entry(row_id);
     aircraft_entry.setData(getEntryData(aircraft_entry.getPosition()));
     return aircraft_entry;
 }
 
-AFlightEntry ADatabase::getFlightEntry(RowId row_id)
+AFlightEntry ADatabase::getFlightEntry(RowId_t row_id)
 {
     AFlightEntry flight_entry(row_id);
     flight_entry.setData(getEntryData(flight_entry.getPosition()));
@@ -497,7 +497,7 @@ const QStringList ADatabase::getCompletionList(ADatabaseTarget target)
 }
 
 const
-QMap<ADatabase::QString, ADatabase::RowId_t> ADatabase::getIdMap(ADatabaseTarget target)
+QMap<QString, RowId_t> ADatabase::getIdMap(ADatabaseTarget target)
 {
     QString statement;
 
@@ -573,7 +573,7 @@ int ADatabase::getLastEntry(ADatabaseTarget target)
     }
 }
 
-QList<int> ADatabase::getForeignKeyConstraints(int foreign_row_id, ADatabaseTarget target)
+QList<RowId_t> ADatabase::getForeignKeyConstraints(RowId_t foreign_row_id, ADatabaseTarget target)
 {
     QString statement = "SELECT ROWID FROM flights WHERE ";
 
@@ -610,12 +610,12 @@ QList<int> ADatabase::getForeignKeyConstraints(int foreign_row_id, ADatabaseTarg
     return row_ids;
 }
 
-APilotEntry ADatabase::resolveForeignPilot(int foreign_key)
+APilotEntry ADatabase::resolveForeignPilot(ForeignKey_t foreign_key)
 {
     return aDB->getPilotEntry(foreign_key);
 }
 
-ATailEntry ADatabase::resolveForeignTail(int foreign_key)
+ATailEntry ADatabase::resolveForeignTail(ForeignKey_t foreign_key)
 {
     return aDB->getTailEntry(foreign_key);
 }
--- a/src/database/adatabase.cpp
+++ b/src/database/adatabase.cpp
@@ -496,7 +496,8 @@ const QStringList ADatabase::getCompletionList(ADatabaseTarget target)
     return completer_list;
 }
 
-const QMap<QString, int> ADatabase::getIdMap(ADatabaseTarget target)
+const
+QMap<ADatabase::QueryResult_qstr, ADatabase::QueryResult_i> ADatabase::getIdMap(ADatabaseTarget target)
 {
     QString statement;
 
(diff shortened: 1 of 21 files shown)


src/database/adatabase.cpp
```diff
@@ -496,7 +496,8 @@ const QStringList ADatabase::getCompletionList(ADatabaseTarget target)
     return completer_list;
 }
 
-const QMap<QString, int> ADatabase::getIdMap(ADatabaseTarget target)
+const
+QMap<ADatabase::QueryResult_qstr, ADatabase::QueryResult_i> ADatabase::getIdMap(ADatabaseTarget target)
```
fiffty-50, line 500: `const QMap<QString, int> ADatabase::getIdMap(ADatabaseTarget target)`

src/database/adatabase.cpp
```diff
@@ -523,24 +524,25 @@ const QMap<QString, int> ADatabase::getIdMap(ADatabaseTarget target)
         break;
     default:
         DEB << "Not a valid completer target for this function.";
-        return QMap<QString, int>();
+        return {};  // [G]: Cpp will implicitly create the default map.
```
fiffty-50, line 527: that's ok!

src/database/adatabase.h
```diff
@@ -88,6 +88,23 @@ class ADatabaseError : public QSqlError {
...
+    using PilotName_qstr = QueryResult_qstr;
+    using PilotRowId_i = QueryResult_i;
+
+    using TailRgstr_qstr = QueryResult_qstr;
+    using TailId_i = QueryResult_i;
+
+    using ArprtICAO_qstr = QueryResult_qstr;
+    using ArprtIATA_qstr = QueryResult_qstr;
+    using ArprtName_qstr = QueryResult_qstr;
+    using ArprtId_i = QueryResult_i;
```
fiffty-50, line 105: `using RowId = int` for ArprtId_i, TailId_i, PilotRowId_i
`using PilotName = QString;`
`using AirportICAO = QString;`
`using AirportIATA = QString;`
`using AirportName = QString;`
`using TailRegistration = QString;`

src/gui/dialogues/newflightdialog.h
```diff
@@ -129,11 +136,11 @@ private slots:
...
-    QMap<QString, int> pilotsIdMap;
-    QMap<QString, int> tailsIdMap;
-    QMap<QString, int> airportIcaoIdMap;
-    QMap<QString, int> airportIataIdMap;
-    QMap<QString, int> airportNameIdMap;
+    QMap<PilotName_qstr, PilotRowId_i> pilotsIdMap;
+    QMap<TailRgstr_qstr, TailId_i> tailsIdMap;
+    QMap<ArprtICAO_qstr, ArprtId_i> airportIcaoIdMap;
+    QMap<ArprtIATA_qstr, ArprtId_i> airportIataIdMap;
+    QMap<ArprtName_qstr, ArprtId_i> airportNameIdMap;
```
fiffty-50, line 143:
```cpp
    QMap<PilotName, RowId> pilotsIdMap;
    QMap<TailRegistration, RowId> tailsIdMap;
    QMap<AirportICAO, RowId> airportIcaoIdMap;
    QMap<AirportIATA, RowId> airportIataIdMap;
    QMap<AirportName, RowId> airportNameIdMap;
```

fiffty-50 (changes requested):
I think the general idea here is good, but I think the idea is taken a bit too far in its implementation.

- Adding the underlying type to the alias is unneeded in my opinion. It makes the aliases convoluted and if need be, finding out the type is as simple as hovering the mouse over the alias (in a modern IDE, I don't think we should create ugly code just to cater to somebody wanting to use vim to write cpp code), or just following the alias to its definition.
- The database queries return 'basic' types (let's call qstring a basic type here), so we should keep those types within the database class, instead of creating the middle step 'QueryResult'. This only decreases readability and increases complexity. Database return values are always QVariants, cast to a type depending on the data they hold. Let's give meaning to those results, instead of creating a Matryoshka of aliases.
- Not every integer needs its own name, for example the different ids can be grouped together since they are always referring to the same thing - the row id

I have left suggestions on how I would name the aliases in the comments above.

All in all, I like the initiative, but I would prefer a more simple approach, **less is more**!

KotzaBoss:
In that case we should follow the C/C++ convention for aliases with the `_t` instead of the `_type`.
To indicate that it is indeed an alias.
```c++
using PilotName_t = QString;
using AirportICAO_t = QString;
using AirportIATA_t = QString;
using AirportName_t = QString;
using TailRegistration_t = QString;
```
You may have even see it in qt i believe. Its a convention for aliases. I suppose we can remove the middle query result step, it can be argued that its too verbose but i was trying to be both explicit and a bit proactive because at least my brain just starts to blur all the qstring together after a while. Will edit later this day

src/database/adatabase.h
KotzaBoss: These and the previous comment have a discrepency. Should the RowId or the int be used in the QMap?

src/database/adatabase.cpp
KotzaBoss: Can you give me examples of functions that cast the qvariant to other types because i still believe a QueryResult_t will make the fact that a query occurs visible at a glance

fiffty-50: Pretty much all of them do, here are 3 examples:
ADatabase::getIdMap
```
    while (query.next()) {
        id_map.insert(query.value(1).toString(), query.value(0).toInt());
    }
```
ADatabase::getLastEntry
```
        return query.value(0).toInt();
```
ADatabase::getForeignKeyConstraints
```
    while (query.next()) {
        row_ids.append(query.value(0).toInt());
    }
    return row_ids;
```

[QSqlQuery::value](https://doc.qt.io/qt-5/qsqlquery.html#value) always returns a QVariant, which is then cast to the needed type (in most functions) or directly written to the QMap<QString, QVariant> aka QMap<ColName, ColData> for the entries.

src/database/adatabase.h
fiffty-50: I would go for RowId

fiffty-50:
> In that case we should follow the C/C++ convention for aliases with the `_t` instead of the `_type`.
> To indicate that it is indeed an alias.

That sounds reasonable. I guess this should also be the case for the aliases defined in `declarations.h`?
```
using ColName = QString;
using ColData = QVariant;
using TableName = QString;
using RowId = int;

using TableNames = QStringList;
using RowData = QMap<ColName, ColData>;
using ColumnData = QPair<ColName, ColData>;
using ColumnNames = QStringList;
using TableColumns = QMap<TableName, ColumnNames>;
```

KotzaBoss:
Indeed i will breakup the declarations.h into their respective modules so that we dont have a monolith of aliases

src/database/adatabase.cpp
KotzaBoss: Yes but the thing is that each function casts it to its specific type so QVariant returns from the db but these functions cast it to something "meaningful"
For each of them what is that?

fiffty-50:
I have been trying to look into the _t convention you mentioned and I am getting a little confused. What I am finding is suggesting to **not** use _t for own types, since this convention is reserved in POSIX, and is associated with the STL in c++.

Examples:
[GNU C library manual](https://www.gnu.org/software/libc/manual/html_node/Reserved-Names.html)

>  Some additional classes of identifier names are reserved for future extensions to the C language or the POSIX.1 environment. While using these names for your own purposes right now might not cause a problem, they do raise the possibility of conflict with future versions of the C or POSIX standards, so you should avoid these names. 
> - (...)
> - Names that end with ‘_t’ are reserved for additional type names. 

[The POSIX standard](https://pubs.opengroup.org/onlinepubs/9699919799/xrat/V4_xsh_chap02.html)
B.2.12 Data Types - Defined Types
>  The requirement that additional types defined in this section end in "_t" was prompted by the problem of name space pollution. It is difficult to define a type (where that type is not one defined by POSIX.1-2017) in one header file and use it in another without adding symbols to the name space of the program. To allow implementors to provide their own types, all conforming applications are required to avoid symbols ending in "_t", which permits the implementor to provide additional types. Because a major use of types is in the definition of structure members, which can (and in many cases must) be added to the structures defined in POSIX.1-2017, the need for additional types is compelling.

Some comments I read generally go along the lines of 

>  Whenever I see something_t I know it is something that is in the standard C library (or POSIX) as opposed to something that I or some third-party created.

> As a C heritage the _t (that used to mean "defined via typedef") syntax has been inherited (they're also SUS/POSIX-reserved in the global namespace).
Types added in C++ and not present in the original C language (e.g. size_type) don't need to be shortened.

> Member types are called type or something_type in the C++ standard library. This is readable and descriptive, and the added verbosity is not usually a problem because users don't normally spell out those type names
(...)
The convention is thus restricted to common, vocabulary-type names. The class names of your actual business logic (including std::string) presumably do not warrant such a naming pattern, and I don't think many people would like to have to mangle every type name.
If you will, the _t and _type naming conventions apply primarily to the standard library and to certain aspects of the standard library style, but you do not need to take them as some kind of general mandate.

Could you point me towards some resources confirming the way we use the _t here is indeed correct? The results I am finding so far don't seem to be very keen on using _t

KotzaBoss:
I dont have a source, its from my use of C where it is always for a typedef as mentioned. We can just change it to capital letter _T and be done with it. It will also look consistent with the pascal case we have atm. 
AliasesType_T

src/database/adatabase.cpp
fiffty-50: getIdMap returns the maps we talked about earlier, for example `QMap<PilotName, RowId> pilotsIdMap;`
getLastEntry returns an `int` aka `RowId`
getForeignKeyConstraints returns a `List<int>` aka `List<RowId>`

Let's say you're interested in one of these maps, you look how it is created and get to the database function - wouldn't it be easier to see the types the map is made of, `<QString, int>` than seeing a custom database type, then following that type to see its qstring and int? Since the maps are already named with aliases, having an aliased middle step feels counter-intuitive.

I have looked through the database functions and only one of them, `disconnect` does not query the database. So in a manner of speaking, you are getting a 'QueryResult_t` when you are using the ADatabase class, which seems clear enough. 

Also, having too many custom names obscures the underlying types. I'm all for using aliases to give meaningful names to otherwise convoluted complex data types, but I am not so eager about giving names to basic data types. There are of course arguments to be made, and I think having an int called RowId is great, since it is used like that in many cases and makes it easier to understand what it is for. Nevertheless, I also like to see at first glance what the underlying type is, when dealing with basic tasks.

src/database/adatabasetypes.h
```diff
@@ -19,38 +19,41 @@
...
-using ColumnData = QPair<ColName, ColData>;
-using ColumnNames = QStringList;
-using TableColumns = QMap<TableName, ColumnNames>;
+using TableNames_t = QStringList;
+using RowData_t = QMap<ColName_t, ColData_t>;
+using ColumnData_t = QPair<ColName_t, ColData_t>;
+using ColumnNames_t = QStringList;
+using TableColumns_t = QMap<TableName_t, ColumnNames_t>;
+
+using ForeignKey_t = int;
```
fiffty-50, line 48: ForeignKey == RowId

Foreign keys are used to reference RowIds in another table, but they are the RowIds of these referenced (Foreign) tables, so I would call them RowId as well

src/database/adatabasetypes.h
```diff
@@ -19,38 +19,41 @@
...
+using TailRgstr_t = QString;
+using TailId_t = RowId_t;
+using ArprtICAO_t = QString;
+using ArprtIATA_t = QString;
+using ArprtName_t = QString;
+using ArprtId_t = RowId_t;
+using ColName_t = QString;
+using ColData_t = QVariant;
+using TableName_t = QString;
+using RowId_t = int;
```
fiffty-50, line 40: Let's spell out the words here

`TailRegistration`,`AirportIATA` etc.

fiffty-50:
> We can just change it to capital letter _T
yeah we can do that.

src/database/adatabasetypes.h
KotzaBoss: I find it funny that a pilot whose entire life is abbreviations wants me to un-abbreviate common fraces like registration and airport

src/database/adatabase.cpp
KotzaBoss: If you have forgotten what "IdMap" means  then seeing just `QMap<QString, int>` doesnt help that much. The aliasing starts from the function singature so that it ends in keeping you sane. First you see that it returns eg `QMap<PilotName_T, RowId_T>` and you either remember on the fly what IdMap means or you continue to look at the function body. 

You said it yourself that checking the type is as easy as hovering the mouse. The documentation can have the basic types so that it gives as much info as possible.
```c++
/// \brief Query the db and return pilotname and corresponding rowid as QString and int.
QMap<PilotName_T, RowId_T> getMap(ADatabaseTarget t);
```

src/database/adatabasetypes.h
fiffty-50: I'm not against abbreviations, they are useful! But neither `Arprt` nor `Rgstr` are (common) abbreviations. Those are like writing `QVrnt` instead of `QVariant`.

Abbreviating and omission of vowels are two different things.

Abbreviation is fine, it would be insane to write `AirportInternationalAirTransportAssociation` instead of `AirportIATA`. That's because IATA or ICAO etc. aren't regular words.

But disemvoweling is different (and bad). After all, we're not writing a SMS limited to 160 characters.

src/database/adatabase.cpp
fiffty-50: are we talking about the same thing? I was referring to having a seperate `QueryResult_t` as being unnecessary, naming the maps themselves is fine.

`QMap<QString, int> getIdMap` should return the basic types because we name them later on, once the maps are created, here it would be infeasible, since the return from getIdMap, <QString, int> can be a variety of maps, for example:

```cpp
QMap<PilotName_T, PilotRowId_T> pilotsIdMap;
QMap<TailRegistration_T, TailId_T> tailsIdMap;
QMap<AirportICAO_T, AirportId_T> airportIcaoIdMap;
QMap<AirportIATA_T, AirportId_T> airportIataIdMap;
QMap<AirportName_T, AirportId_T> airportNameIdMap;
```

Adding another intermediate type only obfuscates the underlying type, since then, if you would 'hover your mouse' over `PilotName` (or look it up), it would show you `QueryResult_t` instead of `QString`,  which is the piece of information you're after.

In general, I believe we should only use aliases when there is a good reason to do so. If every `int` and `QString` end up being their own custom type, it makes keeping an overview more difficult, not less (at least for me)

KotzaBoss: We are talking about the same thing. If you see `getIdMap` return `QString` and `int` and take an argument `AdatabaseTarget` how can you at a glance infer that if you pass the pilot target its `PilotName_T`, `PilotRowId_T` and not `PilotCallsign_T`?

There will have to be a couple of logic steps you have to take anyway. Semantically the getIdMap fucntion queries the db based on the target and casts the return variant to string. So it makes sense to be abstractly aliased as `QueryResult_T` (step 1). And after that you see what is aliased as `QueryResult_T`(step 2). All that if need (which will be in the future).
All that obviously is semantic aliasing so anyone taking the returned QMap has a string in their hand. If we wanted *absolute* control of the return type of course we would have to make a custom one but thats unecessary (because again we want to use a string *but* each string has meaning for the program)

I dont know why we have one one such function and we dont overload it which will pretty much bring an end to this discussion. I can do it in 30 minutes because its just breaking up the switch into overloads and calling the proper ones around the program.

fiffty-50: > how can you at a glance infer that if you pass the pilot target its PilotName_T, PilotRowId_T and not PilotCallsign_T?

From there, the only thing you can infer is that you are getting a string. When you look at the map, that's where you see what the data is. But knowing you are getting a string is better than knowing you are getting some `QueryResult_T` that could be anything.
> All that obviously is semantic aliasing so anyone taking the returned QMap has a string in their hand.

Exactly. That's why I prefer having `QString` over an alias, `QueryResult_T` tells me nothing and creating an alias here is obfuscating the fact that it's no black magic we're dealing with here, just a good old string.
> There will have to be a couple of logic steps you have to take anyway.

One more reason to have as little steps as possible. Series of aliases add levels of abstraction that make it impossible to see at a glance what kind of data we are dealing with here.

src/database/adatabasetypes.h
```diff
@@ -21,39 +21,35 @@
...
+using TableColumns_T = QMap<TableName_T, ColumnNames_T>;
 
-using TableNames_t = QStringList;
-using RowData_t = QMap<ColName_t, ColData_t>;
-using ColumnData_t = QPair<ColName_t, ColData_t>;
-using ColumnNames_t = QStringList;
-using TableColumns_t = QMap<TableName_t, ColumnNames_t>;
-
-using ForeignKey_t = int;
+using ForeignKey_T = RowId_T;
```
fiffty-50, line 44: there's really no need to differentiate this one.
```cpp
resolveForeignPilot(ForeignKey_T foreign_key);
resolveForeignTail(ForeignKey_T foreign_key);
```
can just take a `RowId_T`, since that's what it is.

PR #64, opened 2021-01-29
Develop style options rework

### LINKED ISSUE #62: User-facing style options could be more straight-forward
In the settings widget, the user can select a style for the application.

At the moment, this is done using a Drop-Down menu (QCombobox) filled with the available styles on the system, and an additional Checkbox to activate the dark stylesheet.

This leads to a somewhat confusing user experience, since selecting the stylesheet overrides the selection in the Drop-Down menu.

Since the distinction between style and stylesheet is irrelevant to the end user, I would suggest handling this behind the scenes, by integrating the option for the dark stylesheet into the Dropdown Menu as `Breeze-dark` (and `Breeze-light` for the light theme) so that on each platform we can offer the styles that are available on the system, plus the Breeze styles via stylesheets.

fiffty-50:
closes #62
- User-facing style choices unified into the style combo box in the settingswidget, styleCheckBox removed
- Distinction between style and stylesheets are handled in the AStyle class.
- ASettings adjusted accordingly


src/gui/widgets/settingswidget.cpp
```diff
@@ -357,3 +361,47 @@ void SettingsWidget::on_styleComboBox_currentTextChanged(const QString& new_styl
...
+    case Qt::Checked:
+    {
+        ui->fontComboBox->setEnabled(false);
+        ui->fontSpinBox->setEnabled(false);
+        ASettings::write(ASettings::Main::UseSystemFont, true);
+        QMessageBox message_box(this);
+        message_box.setText(tr("The application will be restarted for this change to take effect."));
+        message_box.exec();
+        qApp->quit();
+        QProcess::startDetached(qApp->arguments()[0], qApp->arguments());
```
KotzaBoss, line 401: Im confused as to who knows where to find what font. When you set use system font you read the combobox on the right to write to the ini so its not system font its what the user chooses. You cant find the default with `QFont::defaultFont()` but i dont think its what the systems user has selected. It might be a fallback font for me its Deja Vu.

KotzaBoss (changes requested):
There are alot of ommitted debug statements making tracking down the font selection pretty hard

src/gui/widgets/settingswidget.cpp
fiffty-50: Yeah that's a bit unclear.

When setting or unsetting using system font nothing is written to the combobox, so that the users last choice is saved and displayed to him, so you can easily switch between using the system font and 'your' custom font. You custom font is inactive (disabled / greyed out) but still available to be activated.

If the user never selected a custom font, then a fallback font is written to the ini, in your case Deja-Vu. Once you do select a custom font though, it should be read correctly.

The problems I had with the default font is that it is read at application startup, but if you set a custom font with `QApplication::setFont`, then this becomes the default font and the 'old' default font gets overwritten. That's also the reason that in order to revert to the system font, a restart is required.

I'll see if I can make the logic there a bit more intuitive.

I'll also add a warning for using a stylesheet in combination with a custom font, it's possible but there are a few catches.

src/gui/widgets/settingswidget.cpp
```diff
@@ -433,3 +434,10 @@ bool SettingsWidget::usingStylesheet()
     }
     return false;
 }
+
+void SettingsWidget::on_resetStylePushButton_clicked()
+{
+    DEB << "Resetting style to default...";
+    ui->styleComboBox->setCurrentText(QStringLiteral("Fusion"));
```
KotzaBoss, line 441: dont we already have a mapping to that string?

fiffty-50: right! `AStyle::defaultStyle` should do the trick

PR #71, opened 2021-02-19
First draft of backup widget, moved ui specific commands to backup.ui

### LINKED ISSUE #69: Develop BackupWidget
The purpose of the `BackupWidget` is to primarily enable backing up and restoring the database to and from `backupdir` as in **AStandardPaths**, but should also allow exporting and importing the database to an external path (a Pen Drive, a cloud folder etc.)

So what I have in mind is a widget containing the following:

- 'Backup Database' (or similar) button that creates a copy of the current database and places it in `backupdir`
- A view of available backups in `backupdir`, maybe a **QTableView** or **QListView** presenting the backup dates from which the user can select a backup.
- 'Restore Backup' (or similar) button that copies the selected backup from the list from `backupdir` to `DatabaseLocation`.
- `Delete Backup` to delete an outdated backup, a duplicate or do some 'housekeeping'
- Similar to `Backup` and `Restore`, the option to `Import` and `Export` the database, but instead of writing/reading to/from the `backupdir`, a QFileDialog (or similar) enables the user to choose the Location of the Backup.

Apart from that basic layout, we also need to think about how to verify that the file the user wants to import is valid, so we need to establish some requirements that a *file* must fulfill in order to be acceptable.

Nice to have: The list of available backups can display a few key dates about the backup itself when selecting a line from the QTableView, for example something like:

"Backup from 2021-02-12, 364 Flights with 748:33h Total Time, last Flight: KJFK-EDDF on 2021-01-03"

KotzaBoss:
closes #69
Work on Backup Widget.

### DIFF
--- a/mainwindow.cpp
+++ b/mainwindow.cpp
@@ -73,13 +73,13 @@ MainWindow::MainWindow(QWidget *parent)
 
 
     // check database version (Debug)
-    int db_ver = checkDbVersion();
+    int db_ver = aDB->dbVersion();
     if (db_ver != DATABASE_REVISION) {
-        DEB << "############## WARNING ####################";
+        DEB << "############## WARNING ##############";
         DEB << "Your database is out of date.";
         DEB << "Current Revision:\t" << DATABASE_REVISION;
         DEB << "You have revision:\t" << db_ver;
-        DEB << "############## WARNING ###################";
+        DEB << "############## WARNING ##############";
         QMessageBox message_box(this); //error box
         message_box.setText(tr("Database revision out of date!"));
         message_box.exec();
@@ -179,13 +179,6 @@ void MainWindow::on_actionDebug_triggered()
 }
 
 // Debug
-
-int MainWindow::checkDbVersion()
-{
-    QSqlQuery query("SELECT COUNT(*) FROM changelog");
-    query.next();
-    return query.value(0).toInt();
-}
 
 // not used at the moment
 
--- a/mainwindow.cpp
+++ b/mainwindow.cpp
@@ -73,7 +73,7 @@ MainWindow::MainWindow(QWidget *parent)
 
 
     // check database version (Debug)
-    int db_ver = aDB->dbVersion();
+    int db_ver = aDB->version();
     if (db_ver != DATABASE_REVISION) {
         DEB << "############## WARNING ##############";
         DEB << "Your database is out of date.";
--- a/mainwindow.cpp
+++ b/mainwindow.cpp
@@ -73,7 +73,7 @@ MainWindow::MainWindow(QWidget *parent)
 
 
     // check database version (Debug)
-    int db_ver = aDB->version();
+    int db_ver = aDB->dbVersion();
     if (db_ver != DATABASE_REVISION) {
         DEB << "############## WARNING ##############";
         DEB << "Your database is out of date.";
--- a/mainwindow.cpp
+++ b/mainwindow.cpp
@@ -126,6 +126,9 @@ void MainWindow::connectWidgets()
                      pilotsWidget,   &PilotsWidget::onPilotsWidget_databaseUpdated);
     QObject::connect(settingsWidget, &SettingsWidget::settingChanged,
                      pilotsWidget,   &PilotsWidget::onPilotsWidget_settingChanged);
+
+    QObject::connect(aDB,             &ADatabase::connectionReset,
+                     logbookWidget,   &LogbookWidget::repopulateModel);
 }
 
 /*
--- a/mainwindow.cpp
+++ b/mainwindow.cpp
@@ -129,6 +129,10 @@ void MainWindow::connectWidgets()
 
     QObject::connect(aDB,             &ADatabase::connectionReset,
                      logbookWidget,   &LogbookWidget::repopulateModel);
+    QObject::connect(aDB,             &ADatabase::connectionReset,
+                     pilotsWidget,    &PilotsWidget::repopulateModel);
+    QObject::connect(aDB,             &ADatabase::connectionReset,
+                     aircraftWidget,  &AircraftWidget::repopulateModel);
 }
 
 /*
--- a/src/database/adatabase.cpp
+++ b/src/database/adatabase.cpp
@@ -753,21 +753,27 @@ bool ADatabase::restoreBackup(const QString& backup_file)
 {
     INFO << "Restoring backup from file:" << backup_file;
 
+    QString default_loc = databaseFile.absoluteFilePath();
+
     ADatabase::disconnect();
     QFile backup(backup_file);
-    QFile current_db(databaseFile.absoluteFilePath());
+    QFile current_db(default_loc);
 
-    if (!current_db.remove(databaseFile.absoluteFilePath())) {
-        WARN << current_db.errorString() << "Unable to delete current db file";
+    if (!current_db.rename(default_loc + QLatin1String(".tmp"))) { // move previously used db out of the way
+        WARN << current_db.errorString() << "Unable to remove current db file";
         return false;
     }
 
-    if (!backup.copy(databaseFile.absoluteFilePath()))
+    if (!backup.copy(default_loc))
     {
         WARN << backup.errorString() << "Could not copy" << backup << "to" << databaseFile;
+        // try to restore previously used db
+        current_db.rename(default_loc);
         return false;
     }
 
+    // backup has been restored, clean up the previously moved file
+    current_db.remove();
     INFO << "Backup successfully restored!";
     ADatabase::connect();
     emit connectionReset();
--- a/src/database/adatabase.cpp
+++ b/src/database/adatabase.cpp
@@ -40,9 +40,9 @@ ADatabase::ADatabase()
                    )
 {}
 
-int ADatabase::version() const
+int ADatabase::dbVersion() const
 {
-    return dbVersion;
+    return databaseVersion;
 }
 
 int ADatabase::checkDbVersion() const
@@ -115,7 +115,7 @@ bool ADatabase::connect()
     // Enable foreign key restrictions
     QSqlQuery query(QStringLiteral("PRAGMA foreign_keys = ON;"));
     updateLayout();
-    dbVersion = checkDbVersion();
+    databaseVersion = checkDbVersion();
     return true;
 }
 
--- a/src/database/adatabase.cpp
+++ b/src/database/adatabase.cpp
@@ -33,26 +33,23 @@ QString ADatabaseError::text() const
 
 ADatabase* ADatabase::self = nullptr;
 
-/*!
- * \brief Return the names of a given table in the database.
- */
+int ADatabase::dbVersion()
+{
+    QSqlQuery query("SELECT COUNT(*) FROM changelog");
+    query.next();
+    return query.value(0).toInt();
+}
+
 ColumnNames_T ADatabase::getTableColumns(TableName_T table_name) const
 {
     return tableColumns.value(table_name);
 }
 
-/*!
- * \brief Return the names of all tables in the database
- */
 TableNames_T ADatabase::getTableNames() const
 {
     return tableNames;
 }
 
-/*!
- * \brief Updates the member variables tableNames and tableColumns with up-to-date layout information
- * if the database has been altered. This function is normally only required during database setup or maintenance.
- */
 void ADatabase::updateLayout()
 {
     auto db = ADatabase::database();
@@ -73,7 +70,7 @@ void ADatabase::updateLayout()
 ADatabase* ADatabase::instance()
 {
 #ifdef __GNUC__
-    return self ?: self = new ADatabase();
+    return self ?: self = new ADatabase();  // Cheeky business...
 #else
     if(!self)
         self = new ADatabase();
@@ -86,10 +83,6 @@ ADatabase::ADatabase()
                              absoluteFilePath(QStringLiteral("logbook.db"))))
 {}
 
-/*!
- * \brief ADatabase::sqliteVersion returns database sqlite version.
- * \return sqlite version string
- */
 const QString ADatabase::sqliteVersion()
 {
     QSqlQuery query;
--- a/src/database/adatabase.cpp
+++ b/src/database/adatabase.cpp
@@ -715,3 +715,61 @@ QMap<QString, QString> ADatabase::databaseSummary(const QString &db_path)
 
     return return_values;
 }
+
+/*!
+ * \brief ADatabase::createBackup copies the currently used database to an external backup location provided by the user
+ * \param dest_file This is the full path and filename of where the backup will be created, e.g. 'home/Sully/myBackups/backupFromOpl.db'
+ */
+bool ADatabase::createBackup(const QString& dest_file)
+{
+    INFO << "Backing up current database to: " << dest_file;
+    ADatabase::disconnect();
+    QFile file(databaseFile.absoluteFilePath());
+    DEB << "File:" << file;  // [G]: Check adebug.h got INFO WARN, ... additions and discuss convention of use.
+
+    if (!file.copy(dest_file)) {
+        WARN << "Unable to backup old database.";
+        DEB << file.errorString();
+        return false;
+    }
+
+    INFO << "Backed up old database as:" << dest_file;
+    ADatabase::connect();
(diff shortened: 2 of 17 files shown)


mainwindow.cpp
```diff
@@ -73,7 +73,7 @@ MainWindow::MainWindow(QWidget *parent)
 
 
     // check database version (Debug)
-    int db_ver = aDB->dbVersion();
+    int db_ver = aDB->version();
```
fiffty-50, line 76: We could call the database 'version' revision to more clearly differentiate between the program and the database. When reading database version it makes me think of the sqlite version as in `ADatabase::sqliteVersion()`

How about this?

`ADatabase::sqliteVersion()` becomes `ADatabase::version()`
`ADatabase::version()` becomes `ADatabes::revision()`

 With the docstrings `\brief` outlining what you get from each.

src/gui/widgets/backupwidget.cpp
```diff
@@ -89,21 +89,29 @@ void BackupWidget::fillTableWithSampleData()
...
-    // Second column, would be created by reading the details from each file and creating the description string
-    QList<QStandardItem *> descriptions;
+    // Get summary of each db file and populate lists (columns) of data
+    QList<QStandardItem *> total_flights, total_tails, total_pilots, max_doft, total_time;
     for (const auto &entry : entries) {
-        QStringList summary = aDB->databaseSummary(entry);
-        auto item = new QStandardItem(summary.join(", "));
-        descriptions.prepend(item);  // Make a O(1) op here and deal with sorting later
+        QMap<QString, QString> summary = aDB->databaseSummary(backup_dir.absoluteFilePath(entry));
+        total_flights.prepend(new QStandardItem(summary["total_flights"]));
```
fiffty-50, line 100: Are you using prepend over append for some efficiency reason or is it just personal preference? (I don't mind, just curious)

fiffty-50 (commented):
Coming along nicely!

mainwindow.cpp
KotzaBoss: yeah... i forgot about that. In that case lets just go with dbVersion. the only problem is that the variable and the getter conflict in name. Ill keep the variable databaseVersion to make it verbose since its not going to be typed on its own asidfe from connect and ill keep the function dbVersion

src/gui/widgets/backupwidget.cpp
KotzaBoss: My comment is my reason. Append is an O(n) operation where as prepend is O(1) and i dont see any reason to waste n complexity time. The thing is though it may reverse the sort order that way so we should check it out.
We should check what entryList returns (ascending or descending) and actually put the opossite of what we want.
All of this could have bern avoided if the tableview accepted vector of items but for some reason it doesnt.

fiffty-50: Maybe we can save us the hassle of dealing with reversing - I think Qt's implementation of QList differs from the standard behavior you are expecting - According to the [docs](https://doc.qt.io/qt-5/containers.html#algorithmic-complexity), append and prepend are both O(1).

qlist

The docs for [QList](https://doc.qt.io/qt-5/qlist.html#append) also have an identical comment for append and prepend 
> If this list is not shared, this operation [append or prepend] is typically very fast (amortized constant time), because QList preallocates extra space on both sides of its internal buffer to allow for fast growth at both ends of the list.

KotzaBoss: oooh cmon... QList is a variation of an array not std::list. QLinkedList is literally a list...
Then no reason not to use append

src/gui/widgets/backupwidget.cpp
```diff
@@ -43,28 +46,66 @@ BackupWidget::~BackupWidget()
...
-void BackupWidget::on_tableView_clicked(const QModelIndex &index) {
-    selected = model->item(index.row(), 0);
-    DEB << "Item at row:" << index.row() << "->" << selected->data(Qt::DisplayRole);
+void BackupWidget::on_tableView_clicked(const QModelIndex &index)
+{
+    selectedBackupName = model->item(index.row(), 0);
+    DEB << "Item at row:" << index.row() << "->" << selectedBackupName->data(Qt::DisplayRole);
 }
 
 void BackupWidget::on_createLocalPushButton_clicked()
```
fiffty-50, line 55: When creating a local backup, there is no need to use a QFileDialog. The idea is that with this button, an 'automatic' backup is created in the local backup folder (and is later displayed in the table for easy "one-click" restoration). So the logic would be something like this (similar to `ADataBaseSetup::backupOldData()`)
```cpp
void BackupWidget::on_createLocalPushButton_clicked()
{
    // create the backup name and path
    QString backup_name = QLatin1String("logbook_bak_")
            + ADateTime::toString(QDateTime::currentDateTime(), Opl::Datetime::Backup)
            + QLatin1String(".db");
    QString full_path = AStandardPaths::directory(AStandardPaths::Backup).absoluteFilePath(backup_name);
    
    // create the backup
    aDB->createBackup(full_path); // with some ifs and checks (e.g. if the filename already exists etc..
    // refresh tableview
}
```

src/gui/widgets/backupwidget.cpp
```diff
@@ -43,28 +46,66 @@ BackupWidget::~BackupWidget()
...
 {
-    DEB << "deleting:" << selected->data(Qt::DisplayRole);
+    if(selectedBackupName == nullptr) {
+        INFO << "No backup was selected";
+        return;
+    }
+    DEB << "deleting:" << selectedBackupName->data(Qt::DisplayRole);
 }
 
 void BackupWidget::on_createExternalPushButton_clicked()
```
fiffty-50, line 107: Here I wouldn't open `AStandardPaths::Backup` since this is used to create an external backup to a custom location specified by the user. The Backupdir is for use with internal backups (the ones that get displayed in the tableview), and the user should not put his 'external' backups in this folder, so I would opt for opening the QFileDialog at `QStandardPaths::HomeLocation`

src/gui/widgets/backupwidget.cpp
```diff
@@ -78,6 +119,7 @@ void BackupWidget::on_createExternalPushButton_clicked()
 
 void BackupWidget::on_restoreExternalPushButton_clicked()
```
fiffty-50, line 121: Same here as above with the backup location, I would advocate for opening `QStandardPaths::HomeLocation`

src/testing/adebug.h
```diff
@@ -11,10 +11,26 @@
     #define FUNC_IDENT __func__
 #endif
 
+// CAUTION: qDebug() doesnt print for non-DEBUG builds
 #define DEB qDebug() << FUNC_IDENT << "\n\t"
 #define DEB_SRC DEB
 #define DEB_RAW qDebug() << '\t'
 
+
```
fiffty-50, line 19: I prefer to stick with DEB for Debugging and LOG for console output to the end user for now, keep it simple. LOG can be used for INFO, WARN and CRIT type messages, so no need to add another 3 additional 'levels' of console output. We can look into colour coding different LOG levels, but I wouldn't make it a priority right now.

I think the NOT_IMPLEMENTED macro is a good and useful addition for WIP sections!

src/database/adatabase.cpp
```diff
@@ -715,3 +715,61 @@ QMap<QString, QString> ADatabase::databaseSummary(const QString &db_path)
...
+ */
+bool ADatabase::restoreBackup(const QString& backup_file)
+{
+    INFO << "Restoring backup from file:" << backup_file;
+
+    ADatabase::disconnect();
+    QFile backup(backup_file);
+    QFile current_db(databaseFile.absoluteFilePath());
+
+    if(!backup.isWritable()) {
```
fiffty-50: good idea!

src/database/adatabase.cpp
```diff
@@ -715,3 +715,61 @@ QMap<QString, QString> ADatabase::databaseSummary(const QString &db_path)
...
+ * \param dest_file This is the full path and filename of where the backup will be created, e.g. 'home/Sully/myBackups/backupFromOpl.db'
+ */
+bool ADatabase::createBackup(const QString& dest_file)
+{
+    INFO << "Backing up current database to: " << dest_file;
+    ADatabase::disconnect();
+    QFile file(databaseFile.absoluteFilePath());
+    DEB << "File:" << file;  // [G]: Check adebug.h got INFO WARN, ... additions and discuss convention of use.
+
+    if (!file.copy(dest_file)) {
```
fiffty-50, line 730: I think this function also needs to be able to overwrite (delete, copy) existing files. QFileDialog asks for confirmation if the user selects an existing file as output, so we could implement the same logic as is used in `restoreBackup`

fiffty-50 (commented):
Nice changes on the database function!

Local backups to and from `BackupDir` need not involve a QFileDialog - the internal default backup directory should be used automatically.

Prefer to not split DEB into 4 different outputs, could be looked into later but for now I would like to stick with DEB for all debugging and LOG for logging. Colour coding is a nice idea, could be looked at later for LOG, but not a priority at the moment.

Good to see the progress, things are starting to come along! 👍🏼

src/gui/widgets/backupwidget.cpp
KotzaBoss: Ah i misunderstood the functionality.
But there should be a check for whether the current db has already been backedup?
Or at least a quick prompt to give some feedback of the action and not just silently create it.

src/gui/widgets/backupwidget.cpp
KotzaBoss: WIll implement!

src/testing/adebug.h
KotzaBoss: The idea is that Info Warn Crit are actually part of the application. Its more or less what the user would copy paste as "error" messages for us to figure out. DEB and so qDebug() is literally just that developement debug so while we can push it back the distinction should be established and appropriately used.
For example any necessary info like actions pressed, switching logic due to some condition etc that when the actual program is running is necessary should _not_ be debug. However internal variable values (pointers etc) can be debug

src/database/adatabase.cpp
KotzaBoss: I missed this i think.
So the function should by default overwrite the dest_file?
The assumption is that QFileDialog is our "guard" because it prompts the overwrite?
That means that the user pressing yes will overwrite an existing file where as no means he _will_ put a new file which doesnt exist and just create it
Please confirm my understanding, it kind of flew over my head

src/gui/widgets/backupwidget.cpp
fiffty-50: Yeah that would be a good idea. The question is how we would go about it, maybe we could compare the database summary and prompt the user if `summary(now) == summary(lastBackup)` or we could check if a filename already exists? In that case we would only look at timestamps in a manner of speaking, as compared to looking at database states.

Not quite sure, what do you think?

src/database/adatabase.cpp
fiffty-50: The idea was that QFileDialog prompts the user to confirm if he really wants to select the existing file, which is default behaviour as far as I'm aware. If that's the case, then the function should silently overwrite since confirmation has already been given. But we would of course have to make sure that it is only called if a confirmation has been given. I need to look into what QFileDialog returns when the action is cancelled, let me get back to you on that.

src/testing/adebug.h
fiffty-50: Hmm, since we are building a GUI application here, the vast majority of users is going to run the app from a start menu / app launcher or similar, meaning they never get to see any console output anyway, regardless of whether it's qDebug, qWarning or anything else.

If we are interested in bug / crash reports, those would have to be written to a text file anyway before being sent to us, so that any colourful coding or differentiating between how messages are displayed would be moot.

LOG can provide all we need in this case, which is logging to console (still need to implement the logging to file). One of the reasons why I'm opposed to differentiating it further is that it opens up a whole range of discussions about is this more suitable for WARN or for CRIT or for INFO. Or even if not discussions, it's still another mental step when producing console out, plus it makes the code more convoluted to read and write.

I'm a big fan of the [KISS principle](https://en.wikipedia.org/wiki/KISS_principle) and try to keep it as a guiding principle for this project, so let's keep it simple stupid ;)

src/gui/widgets/backupwidget.cpp
KotzaBoss: there is the case that the current db is not backed up in a file so thats another edge case...
These kind of problems are solved with hashing. Each db is hashed and thats how you compare it. Im sure Qt offers hashing functions. So in essence
```c++
int hash_of_running_db = hashFunc(ADatabase::database());
int hash_of_backup = hashFunc(ADatabase::readDb(backup_file));
if(hash_of_running_db == hash_of_backup) {
   // dont backup
}
```
The problem when hashing is simply agreeing upon what to hash and it makes sense to be the contents

src/testing/adebug.h
KotzaBoss: Dealing with the color is engineering not a problem.
image

At the same time there are levels of output the application outputs. Logging everything as LOG is _too_ simple. Yes keep it simple but not simpler should be the moto.

Check any logging api you can find they all categorise output. Many in much more extensive ways than info warn crit. As they should bcause thats how a program runs. 

The discussion of what is what (info or warn) is unavoidable and should follow some convention based on the devs' criteria.

src/gui/widgets/backupwidget.cpp
fiffty-50: Yeah that sounds good, we could look into [QCryptographicHash](https://doc.qt.io/qt-5/qcryptographichash.html).

But I think we should keep the option to create another backup of an equal database around. Some people don't care if it's the same, they maybe want to do let's say a weekly backup for their OCD so that they don't feel like their last backup is 'too old'. So we could do something like

```cpp
[...]
if(hash_of_running_db == hash_of_backup) {
   // QMessageBox saying "The current db is already backed up as <filename>. Are you sure you want to create another backup?"
    if (reply == QMesageBox::Yes)
        // well, okay then... let's create another file
    else
        // don't backup
}

```

src/testing/adebug.h
fiffty-50: Most logging API's have different log levels to differentiate between the urgency of the messages that are being displayed, with the rationale being that higher importance messages should catch the attention of the user/sysadmin etc. This means that in a text file, or a status console, messages are printed in different colours, or with different headers, to make them differentiable. This makes a lot of sense in those scenarios, obviously.

However, in a GUI application we need to handle this differently. Anything that should be a "Warning" that should let's say catch the attention of the user (say, an invalid input) has to be displayed in the GUI, as a `QMessageBox` for example, since the end user doesn't see the console output - he's not even running a terminal. What good does a console warning about an invalid input do if the user never gets to see it?

That being said, for hunting down bugs and pinpointing where something went wrong, logging milestones is definitely helpful. I will also agree that having normal 'milestones' like a new flight being logged successfully being differentiated from 'non-normal' returns in functions is a sensible thing, so having different LOG levels is something I am willing to implement.

However, let's take this step by step and not get distracted by it now. I will look into setting up logging and we can discuss the implementation once you are finished with the backup widget. Let's stick with DEB and LOG for now and I will have something ready by the time the backup widget is finished!

src/gui/widgets/backupwidget.cpp
KotzaBoss: simple enough, will implement in the following commits

KotzaBoss:
write down the function you want to generate the filename string and ill use it

fiffty-50:
> 
> 
> write down the function you want to generate the filename string and ill use it
```cpp
#include "src/functions/adatetime.h"
/*!
 * \brief creates a filename to be used for creating a backup file. Returns the
 * full (clean) path with the standard backup directory, e.g.:
 * "/home/john/.local/share/opl/openPilotLog/logbook_backup_2021_03_24.db"
 * "C:/Users/john/AppData/Roaming/opl/openPilotLog/backup/logbook_backup_2021_03_24.db"
 */
QString BackupWidget::backupname()
{
    const QString backup_name = QLatin1String("logbook_backup_")
                + ADateTime::toString(QDateTime::currentDateTime(), Opl::Datetime::Backup)
                + QLatin1String(".db");
    return AStandardPaths::asChildOfDir(AStandardPaths::Backup, backup_name);
}
```

PR #75, opened 2021-04-20
Develop logging debugging

fiffty-50:
Added a new System for handling Debugging, Logging and User Notification, using a qMessageHandler  to intercept and redirect calls to qDebug, qInfo, qWarning and qCritical

These messages are then formatted and written (depending on their message type) to console and/or log file and/or the user in the GUI. Added a logs folder to AStandardPaths, the log files are rotated (last 10 logs are kept).

Logging_and_Debugging

There are a number of convenience macros available - these are mostly unchanged. The available macros are:
- DEB - Messages go to Console out. Normally, they would not go into the log file, but for early testing, logging debug to the log file is also enabled
- LOG - Logging of important application milestones. These are logged to console and log file, but not displayed to the user.
- INFO - Written to console, log file and displayed to the user in the GUI. Use INFO if the information is important enough that the user should be made aware of this. If this is not required, use LOG.
- WARN - Same as INFO, but displays a Warning instead of an information to the user
- CRIT - Same as WARN, but with a critical Warning message.

Code is documented but some more detailed docs are in the works


KotzaBoss:
The debugging "concept" looks good but there are somethings i can comment on.
- Color should be limited to the logging header `[HEADER] ...` otherwise its just too much.
- I dont see the reason to have opl in the header
- the nice indentation in console output should be a postprocess of the text. Meaning that the strings for each message should be just a normal paragraph with only `\n`. The debugging utility should then change the new line to `\t\n` making it indented if necessary. Otherwise the output will be abit chaotic between console and files uncessesarily. The conversion is trivial:
  ```c++
  LOGGING_LEVEL << "some, potentially multiline, warning";
  // In the logging output logic
  msg.replace("\n", "\t\n");
  ```

